PR #82, opened 2019-07-02
New extractors for v 5.0.0 plus retrocompatibility + tests

nbonacchi:
didn't flake :P


nbonacchi:
Travis will fail because of flake... wanna bet?

oliche:
So you forked the groom and load encoder functions which is good since we **want** to freeze the code of previous versions to make the extraction reproducible.

Just need to flake and let's go with the merge.

PR #102, opened 2019-09-26
cca functions from code camp

themattinthehatt:


oliche:
The build is broken: https://travis-ci.org/int-brain-lab/ibllib/builds/589708086?utm_source=github_status&utm_medium=notification
I think you need to  set test as a module and flakify before merging ?

PR #105, opened 2019-09-27
Create dim_reduction.py

mschart:


oliche:
Michaël, could you please flake ? I think there is only one thing.
Also it seems there is a huge example in the main function, with hard-coded paths do you want to commit this someplace else ?
Also you guys at the code camp talked about cross-validation is this already a feature, I can't see it anywhere.

On my end I can handle the extra dependencies for ibllib.

PR #126, opened 2020-02-17
look for ephys file fix

jaib1:
updated an example script


jaib1:
ok cool I will update this

PR #134, opened 2020-02-26
Edited tutorial on one.search_terms()

GaelleChapuis:
so it's clear list in tutorial is non-exhaustive


oliche:
moving to a docs branch

PR #143, opened 2020-03-06
Brainbox examples

jaib1:
Updated after @oliche suggestions


jaib1:
@oliche does this look good to you?

PR #150, opened 2020-04-09
unified env instructions

jaib1:


jaib1:
ok, I moved instructions over to 'iblenv', and here I've just deleted the extraneous .yaml files and made minor updates to the README. let me know what you think

PR #158, opened 2020-04-24
Raise an informative error on token request 500 response

k1o0:
On my machine it wasn't raising an exception when the response was 500.  This change ensures we get an informative error message when Alyx is down.


oliche:
Absolutely. Thanks.

PR #161, opened 2020-05-14
Refactoring of ONE and AlyxClient to use singleton pattern for Alyx connection

nbonacchi:
Here it is, I copied my slack post here:

> By definition singletons allow only one instance do this would create a problem if we'd need to two of them e.g. an ONE obj that connects to the dev and another that connects to the test DB. Coincidentally modules in python are all singleton objects, this means that importing any module multiple times means importing it only once, this is apparent when importing module you are developing then changing a couple of lines and having to close tje python interpreter to access the changes as reimporting it does not really reimport it.
> No changes in the AlyxClient itself, just the addition of the import oneibl.params at the top to be able to create the singleton object at the end of the webclient module.
> ```python
> # Get default params from local file
> _par = oneibl.params.get(silent=False)
> # Try creating singleton AlyxClient object
> try:
>     alyx_client = AlyxClient(base_url=_par.ALYX_URL, username=_par.ALYX_LOGIN, password=_par.ALYX_PWD)
>     print('Connected to ' + _par.ALYX_URL + ' as ' + _par.ALYX_LOGIN,)
> except requests.exceptions.ConnectionError:
>     raise ConnectionError("Can't connect to " + _par.ALYX_URL + '. \n' +
>                             'IP addresses are filtered on IBL database servers. \n' +
>                             'Are you connecting from an IBL participating institution ?')
> # Init connection to Globus if needed
> 
> ```

> So now the first time anyone imports the webclient module in any of the possible import patterns that exist in python, the singleton gets instantiated  with the "default" params, i.e. using username/password/base_url from the params file.

> 
> I added a new AlyxClient instance param `self._obj_id = id(self)` for testing.
> Interface with AlyxClient is exactly the same, i.e if you want your own AlyxClient just import the class and make an instance as usual.
> The webclient module now has an extra thing you can import from it:
> `from oneibl.webclient import alyx_client`  This is the singleton object.
> All previous tests pass, and I'm writing the tests for making sure that the id of the singleton is reused in the user case and its new in the dev cases  (when AlyxClient or the ONE object instances have ben initiated with any username/password/base_url param
> in ONE the changes are also minimal:
> ```python
> class ONE(OneAbstract):
>     def __init__(self, username=None, password=None, base_url=None, silent=False):
>         # get parameters override if inputs provided
>         self._par = oneibl.params.get(silent=silent)
>         self._par = self._par.set('ALYX_LOGIN', username or self._par.ALYX_LOGIN)
>         self._par = self._par.set('ALYX_URL', base_url or self._par.ALYX_URL)
>         self._par = self._par.set('ALYX_PWD', password or self._par.ALYX_PWD)
>         # By default set internal AlyxClient is the module singleton object
>         self._alyxClient = wc.alyx_client
>         # If any parameters are entered create new ALyxClient object
>         if username or password or base_url:
>             try:
>                 self._alyxClient = wc.AlyxClient(username=self._par.ALYX_LOGIN,
>                                                  password=self._par.ALYX_PWD,
>                                                  base_url=self._par.ALYX_URL)
>                 print('Connected to ' + self._par.ALYX_URL + ' as ' + self._par.ALYX_LOGIN,)
>             except requests.exceptions.ConnectionError:
>                 raise ConnectionError("Can't connect to " + self._par.ALYX_URL + '. \n' +
>                                     'IP addresses are filtered on IBL database servers. \n' +
>                                     'Are you connecting from an IBL participating institution ?')
>             # Init connection to Globus if needed
>     @property
>     def alyx(self):
>         return self._alyxClient
> ```
> I removed my previous implementation and basically I now just create the ONE instance AlyxClient by pointing to the singleton in the module. So now all instances of ONE will point to the same singleton object.
> If one however passes any of the input parameters username/password/base_url when instantiating ONE the ONE instance will create a new AlyxClient with the provided params.
>

### DIFF
--- a/alf/io.py
+++ b/alf/io.py
@@ -425,7 +425,7 @@ def _regexp_session_path(path_object, separator):
     """
     Subfunction to be able to test cross-platform
     """
-    return re.search(rf'/\d\d\d\d-\d\d-\d\d/\d\d\d',
+    return re.search(r'/\d\d\d\d-\d\d-\d\d/\d\d\d',
                      str(path_object).replace(separator, '/'), flags=0)
 
 
--- a/ibllib/ephys/ephysqc.py
+++ b/ibllib/ephys/ephysqc.py
@@ -249,7 +249,7 @@ def unit_metrics_ks2(ks2_path=None, m=None, save=True):
 
     if save:
         #  the file name contains the label of the probe (directory name in this case)
-        r.to_csv(ks2_path.joinpath(f'cluster_metrics.csv'))
+        r.to_csv(ks2_path.joinpath('cluster_metrics.csv'))
 
     return r
 
--- a/ibllib/pipes/iblpipe.py
+++ b/ibllib/pipes/iblpipe.py
@@ -266,8 +266,8 @@ def run_task(name, session_dir, dependencies=()):
     return 0
 
 
-def _count(l):
-    return (l)
+def _count(invar):
+    return (invar)
 
 
 class Pipeline:
--- a/oneibl/one.py
+++ b/oneibl/one.py
@@ -89,7 +89,7 @@ SEARCH_TERMS = {  # keynames are possible input arguments and values are actual 
 
 
 class ONE(OneAbstract):
-    def __init__(self, username=None, password=None, base_url=None, silent=False):
+    def __init__(self, username=None, password=None, base_url=None, silent=False, printout=True):
         # get parameters override if inputs provided
         self._par = oneibl.params.get(silent=silent)
         self._par = self._par.set('ALYX_LOGIN', username or self._par.ALYX_LOGIN)
@@ -110,7 +110,8 @@ class ONE(OneAbstract):
                                       "Are you connecting from an IBL participating institution ?")
             # Init connection to Globus if needed
         # Display output when instantiating ONE
-        print(f"Connected to {self._par.ALYX_URL} as {self._par.ALYX_LOGIN}",)
+        if printout:
+            print(f"Connected to {self._par.ALYX_URL} as {self._par.ALYX_LOGIN}",)
 
     @property
     def alyx(self):
@@ -373,7 +374,7 @@ class ONE(OneAbstract):
         for ind, tab in enumerate(_ENDPOINTS):
             if tab == table:
                 field_name = table_field_names[_ENDPOINTS[tab]]
-                full_out.append(self._alyxClient.get('/' + _ENDPOINTS[tab]))
+                full_out.append(self.alyx.get('/' + _ENDPOINTS[tab]))
                 list_out.append([f[field_name] for f in full_out[-1]])
         if verbose:
             pprint(list_out)
@@ -585,6 +586,31 @@ class ONE(OneAbstract):
                            number=session_path.parts[-1])
         # Return the uuid if any
         return uuid[0] if uuid else None
+
+    def get_details(self, eid, full=False):
+        """ Returns details of eid like from one.search, optional return full
+        session details.
+        """
+        # If eid is a list of eIDs recurse through list and return the results
+        if isinstance(eid, list):
+            details_list = []
+            for p in eid:
+                details_list.append(self.get_details(p, full=full))
+            return details_list
+        # If not valid return None
+        if not is_uuid_string(eid):
+            print(eid, " is not a valid eID/UUID string")
+            return
+        # load all details
+        dets = self.alyx.rest("sessions", "read", eid)
+        if full:
+            return dets
+        # If it's not full return the normal output like from a one.search
+        det_fields = ["subject", "start_time", "number", "lab", "project",
+                      "url", "task_protocol", "local_path"]
+        out = {k: v for k, v in dets.items() if k in det_fields}
+        out.update({'local_path': self.path_from_eid(eid)})
+        return out
 
 
 def _validate_date_range(date_range):
--- a/oneibl/one.py
+++ b/oneibl/one.py
@@ -104,12 +104,13 @@ class ONE(OneAbstract):
                 self._alyxClient = wc.AlyxClient(username=self._par.ALYX_LOGIN,
                                                  password=self._par.ALYX_PWD,
                                                  base_url=self._par.ALYX_URL)
-                print(f"Connected to {self._par.ALYX_URL} as {self._par.ALYX_LOGIN}",)
             except requests.exceptions.ConnectionError:
                 raise ConnectionError(f"Can't connect to {self._par.ALYX_URL}.\n" +
                                       "IP addresses are filtered on IBL database servers. \n" +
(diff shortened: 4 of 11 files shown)


nbonacchi:
Humm so one of the tests requires a valid username and password in the local params file
It tests the ONE constructor with partial inputs and uses local .one_params values for the non inputted params. Meaning the local params should exist and that user should have acces to the prod or dev DB...

nbonacchi:
Humm OK so while there might be a problem with the users in the default and test db is not for travis, its for everyone else that wants to run the tests
Travis is failing because of the silent setup which is called bu instantiating ONE the import actually triggers the connection of the singleton with no params, fixable... 2 secs

nbonacchi:
OK problem solved now I have to figure out why my flake8 passes and Travis' doesn't 
?

nbonacchi:
Ah the dev branch is broken...

flake8 .
./ibllib/ephys/ephysqc.py:252:36: F541 f-string is missing placeholders
./ibllib/pipes/iblpipe.py:269:12: E741 ambiguous variable name 'l'
./alf/io.py:428:22: F541 f-string is missing placeholders
./tests/ibllib/test_io.py:191:17: E741 ambiguous variable name 'l'
./tests/ibllib/test_io.py:227:37: E741 ambiguous variable name 'l'
./tests/ibllib/test_io.py:227:68: E741 ambiguous variable name 'l'
./tests/ibllib/test_pipes.py:288:50: E225 missing whitespace around operator

nbonacchi:
Update on AlyxClient Singleton implementation:

Singleton AlyxClient basically uncouples ONE from the connection to the DB 
**IMPORTANT**: ONE is not a singleton, the AlyxClient is. The behavior of ONE is the same so creating an object with no inputs will use the .one_params.json credentials, so all objects and modules that create an ONE instance need to have a way of overloading the ONE object creation with the specific ONE instance required for the current usage.
The only case where this might be a problem is in 2nd order modules/objects that depend on an instance of ONE.

Here is one case where we might develop a module that uses one = ONE() as a global for the module and we want to write a test for it.

Lets say I have a **module.py** file that defines a list function the returns the module level one.list method:
```
from oneibl.one import ONE

one = ONE()


def list(eid):
    return one.list(eid)
 

```
To test this module.py I have a **test_module.py** that will use a known eid from our test DB and check the output of the list function.
```
import unittest
import module
from oneibl.one import ONE

one = ONE(base_url='https://test.alyx.internationalbrainlab.org', username='test_user', password='TapetesBloc18')

class TestModule(unittest.TestCase):
    def setUp(self):
        self.one = one
        seld.eid = 'cf264653-2deb-44cb-aa84-89b82507028a'  # session only on test db (subj clns07390)

    def test_list(self):
        expected = ['some', 'list', 'of', 'dstypes']
        queried = module.list(self.eid)
        self.assertTrue(all([x in expected for x in queried]))
        

```
Now, if I run this locally it will fail, and the reason it fails is because this subject only exists in the test DB and the function **module.list()** uses **module.one** obj that was created on import. Because it was created using no inputs it will use my local .one_params.json credentials.
(**NB:** Travis is fine as ONE is setup silently with the test parameters meaning the "default" AlyxClient authenticates with the test DB)

If you look carefully ONE in the above example test, is not really doing anything unless I have a way to pass it to the function I want to test. For objects it's OK usually there is a way to pass an ONE instance to a constructor (e.g. RegistrationClient, Patcher, ... ). For module imports however we have to use a different trick, i.e., redefine the module variable that is used by its functions. We can do this simply by adding this line after we declared what ONE we want to use: `module.one = one` this will set the module one instance which is used by the module.list function to be out test ONE instance and not the default one.
This works because importing a python module basically means running the first indentation level without looking at the contents until the functions are called. 

**SUMMARY**: Objects that depend on ONE should have a way of being constructed using a custom ONE instance (this is already the case AFAICT). The module level import and instantiation of an ONE object should be avoided unless you remember to patch the one instance inside the module when writing tests for it. 

- **USERS**: Nothing changes: they use ONE with the default .one_params.json configuration. ONE will use the singleton alyx_client from the AlyxClient module. It will be faster and more efficient.

- **ADVANCED USERS**: People can use ONE to create their own scripts so in addition to the USERS behavior, there is the possibility of locally instantiating an object that bypasses the AC singleton so people can call myone = ONE(base_url=tototiti) and have an object that uses those credentials to a specific DB. dealing with two connections to the DB is up to the user/dev.

- **DEVS**: When developing a module or an object if the module/object depends on ONE remember to allow inputting a different ONE instance for testing. 

> In any case this made me think that we cannot have screwups at this level, i.e. a user that with the best of the intentions wants to do stuff on the prod DB and really just deletes a bunch of entries. Right now ONE has exposed only "read" methods (list, search, load) but as we're increasingly writing examples that use the one.alyx.rest() interface the possibility of someone screwing up is increased.
> **Conclusion**: We ought to really fix the read/write access to Alyx thing so we can control write access, i.e. POSTS are allowed but PUT/PATCH and DELETE are restricted. (I've added this to our Alyx backlog)

nbonacchi:
**UPDATE 2:**
Implemented singleton using metaclass pattern

**UniqueSingletons** class will return the instance of any class that was initialized using the same parameters, it checks the class name and the *args/**kwargs, i.e. For either AlyxClient or ONE, if the **username** or **base_url** changes it will create a new singleton class and store it in the metaclass attribute **_instances** so `webclient.UniqueSingletons._instances` has a record of all the instantiated classes.
All tests pass and this should work exactly as before but more efficient

I considered several alternative implementations:
- Using class **@decorators**: implementation would be basically the same but it would hide the created instances under a variable of the decorator function. I'm not sure we could keep this in memory and storing "important" variables in a local scope of a function did not sound like a good idea.

- Using a **BaseClass**: I discarded this for 2 reasons, it would require the AlyxClient to inherit from the UniqueSingletons class AND BaseClass. I avoid multiple inheritances because, multiple inheritances... but most importantly the constructor of the AlyxClient object would re-run.

- Using the **`__new__`** special method: The `AlyxClient.__new__` special method, runs before `__init__ `and that's OK but it has the same problem of the BaseClass pattern i.e. the `AlyxClient.__init__` constructor would run right after, this means the object would be the same in terms of pointer and memory but it would be rewritten, so it would re-authenticate to the server which defeats the purpose.  

**Finally**:
ONE has no changes in terms of behavior `ONE.__init__` behaves as usual i.e. it creates the  AlyxClient object. The AlyxClient is a bit smarter and will return the relevant instance instead of creating one if it exists already. It's super fast as it skips the authentication part in case the specific AlyxClient object was already created. 
I think we can now add an offline mode to ONE that basically skips the AC creation all together, atm it's the only thing the `ONE.__init__` does is deal with the AC, we would have to figure out a way of building an AlyxClient object without authenticating to give back to ONE otherwise `one.alyx.**something**` would not exist and it defeats the purpose for the "coding in a plane" scenario...

But I'll leave that for the next PR...

Issue #24, opened 2018-10-04
one.info  : useless - Delete ?


GaelleChapuis:
one.info > Seems useless, no extra information. Delete entirely?

oliche:
Yup.

PR #203, opened 2020-09-07
Sglx fix

mayofaulkner:
AP and LFP gain info in metadata has 384 entries regardless of the number of actual channels used for recording. Led to error in issue [#24](https://github.com/int-brain-lab/iblenv/issues/24) in iblenv. Fixed by reading in the number of recording channels from metadata and indexing gain by this. 

All local tests have passed, but I haven't run integration tests


oliche:
This should probably be covered by a unit test as we want this spike glx module to be very robust, and this is an example of a kind of acquisition configuration we've never seen before. The idea is to add the small meta-data file and modify it so it is compatible with the mock ephys bin file already covered in the tests.

PR #211, opened 2020-09-17
compute training status using ONE

mayofaulkner:
I have added a module in brainbox.behaviour called training that allows you to compute the training status of a subject based on either the last three sessions that have data, or from a specified date. It uses the ibl_pipeline.utils.psychofit, so i'm not sure if brainbox is the best place for it as we don't really want brainbox relying in datajoint. Perhaps we could move that module into brainbox and have iblpipeline import from brainbox instead??


k1o0:
Looks good! Shall we go over it this week?

brainbox/behavior/training.py
```diff
@@ -0,0 +1,517 @@
...
+        if len(ephys_sess_dates) == 0:
+            assert(np.all(np.array(task_protocol) == 'biased'))
+            if criterion_ephys(info.psych_20, info.psych_80, info.n_trials, info.perf_easy,
+                               info.rt):
+                status = 'ready4ephysrig'
+            else:
+                status = 'trained 1b'
+
+        elif len(ephys_sess_dates) < 3:
+            assert(np.all(np.array([date in trials for date in ephys_sess_dates])))
```
k1o0, line 235: A little clearer w/o numpy `assert all(date in trials for date in ephys_sess_dates)`

k1o0 (commented):
Looks great! Only two suggestions: 1. remove that if-else statement and 2. maybe add the psychofit example notebook to the repo as well?

k1o0:
Feel free to delete the branch if you're done with it

PR #220, opened 2020-10-06
Create saturation_detection.py

mschart:


oliche (changes requested):
Regarding the format this is more an example than a function.
Params are hard-coded, one is instantiated blindly in the function and we should use the logger instead of printing. 
 Maybe it should go in examples ? Or not even on the repository ? 

Regarding the usage, what is the validity of inferring the saturation on post-spike-sorted data ? I though the goal here was to use the raw AP-files and compute this QC as part of the extraction.

We probably need to discuss

PR #221, opened 2020-10-06
load_object and load_dataset methods

k1o0:


oliche:
Related issues
https://github.com/int-brain-lab/ibllib/issues/222
https://github.com/int-brain-lab/iblenv/issues/60

PR #223, opened 2020-10-12
Partial raw data download via HTTP content range

rossant:
Test with:

```python
from oneibl.one import ONE
o = ONE()
arr = o.download_raw_partial(
'http://ibl.flatironinstitute.org/hoferlab/Subjects/SWC_043/2020-09-23/001/raw_ephys_data/probe00/_spikeglx_ephysData_g0_t0.imec0.ap.e9fabe60-2075-41af-ba5a-b126ca965f88.cbin',
'http://ibl.flatironinstitute.org/hoferlab/Subjects/SWC_043/2020-09-23/001/raw_ephys_data/probe00/_spikeglx_ephysData_g0_t0.imec0.ap.8f81ad37-a9ea-4406-9de4-46531d83374b.ch',
10, 13)
print(arr)
```

### DIFF
--- a/oneibl/one.py
+++ b/oneibl/one.py
@@ -1,5 +1,6 @@
 import abc
 import concurrent.futures
+import json
 import logging
 import os
 from pathlib import Path, PurePath
@@ -180,6 +181,7 @@ class OneAbstract(abc.ABC):
         # if empty in parameter file, do not allow and set default
         if not cache_dir:
             cache_dir = str(PurePath(Path.home(), "Downloads", "FlatIron"))
+        assert cache_dir
         return cache_dir
 
     def _make_dataclass_offline(self, eid, dataset_types=None, cache_dir=None, **kwargs):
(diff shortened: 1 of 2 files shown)


oliche:
Impressive !
The download HTTP add-on is great, glad you didn't have to fork this.

The only comment is about the overall architecture, it would be nice to be able to do this without instantiating ONE, no ?
We could keep this one.method, that calls a spikeglx.Reader static method of function instead ? Or even as part of mtscomp (the chunking etc...) ?

rossant:
I'll move the chunking operation to mtscomp (which I had already implemented in a more limited manner a few days before! I'll merge the two in a single function)

Regarding ONE, the reason I did it like this was that it directly provided the login and password to download the data. Another thing is that it may be nice to alter the signature and just specify the eid of the session, rather than providing the explicit URLs. For this I guess we need ONE.

PR #224, opened 2020-10-12
Feature/server scripts

nbonacchi:
Added server script logic to find, move, and rename badly transferred passiveCW sessions to their corresponding ephys sessions.
Will make a PR with the CLI and sh file to be added to the cron jobs in iblscripts...

Implemented fake methods for creating the file footprint of properly transferred ephys session and badly transferred ones. This way the tests for this script can be unittested directly in ibllib. 
These methods live in tests/fixtures/utils.py and might be useful for others.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 99
-ignore = W504, E266
+ignore = W504, W503, E266
 exclude =
     .git,
     __pycache__,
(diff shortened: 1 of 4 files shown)


nbonacchi:
For some reason `test_create_alyx_probe_insertions` fails both locally and on Travis... has the test DB state changed? Does this test clean up after himself? Will check tomorrow after test DB is rebuilt.

k1o0:
> For some reason `test_create_alyx_probe_insertions` fails both locally and on Travis... has the test DB state changed? Does this test clean up after himself? Will check tomorrow after test DB is rebuilt.

I fixed the issue on my PR, do `git cherry-pick a735e68d77fb85f3100d85c29b8575f22526c6a2`.

nbonacchi:
yay it works :) thanks @k1o0

PR #232, opened 2020-10-27
IblError now inherits Exception instead of BaseException

berkgercek:
Inheriting BaseException instead of Exception in a custom error leads to undesirable behavior, especially in the context of `try/except` statements. `try/except` can't catch BaseException-derived errors (At least in 3.8, haven't tried in older versions of py) and leads to a lot of head-scratching. Unless there's a good reason not to, we should follow the Python guideline on this:

https://docs.python.org/3.8/library/exceptions.html#BaseException

Which recommends that you don't use BaseException but rather Exception for custom errors.

### DIFF
--- a/ibllib/exceptions.py
+++ b/ibllib/exceptions.py
@@ -1,4 +1,4 @@
-class IblError(BaseException):
+class IblError(Exception):
     explanation = ''
 
     def __init__(self, *args):


nbonacchi:
should we add a unittest for custom error classes?

k1o0:
> should we add a unittest for custom error classes?

If you'd like to add one I'll hold off from merging.  I'm not sure there's much to test at the moment...

nbonacchi:
> > should we add a unittest for custom error classes?
> 
> If you'd like to add one I'll hold off from merging. I'm not sure there's much to test at the moment...

Depends on the urgency... I have to leave now, but if we don't want to be caught in this again a unittest will enforce it, plus I'm unclear if it's an issue after the 3.8 migration or if it had always been broken and no one noticed until now...

PR #239, opened 2020-11-09
Path offline

k1o0:
eid from path and vice versa now methods of OneOffline


k1o0:
Here's are the methods I added to OneAbstract.  From now on we should implement as much as possible in the super class, then the Alyx class can hit the database before/after calling the offline method?

PR #240, opened 2020-11-09
Create docs_access_DLC.py

mschart:


oliche:
LGTM
One comment: usually variables names are lower case `xys` and we reserve the capital letters for classes  `my_class = MyClass`
It's quite confusing for us, and I think it's a good habit to pass on users, especially in the examples.

I'll merge this as is, but please try to avoid this in the future (it sounds nitpicking, but in the long run it's a good convention).

PR #245, opened 2020-11-16
Various changes to the GLM functionality

berkgercek:
Primarily includes the addition of pytorch-accelerated fitting of GLMs. Will need to have pytorch as a dependency. Working now on making sure that is the case.


berkgercek:
Alright, so this pull makes pytorch a requirement of ibllib and adds torch-accelerated GLM functionality. It's crucial to my project (stepwise regression GLM fits are not possible without it) and adds a substantial speedup. If we want to GPU-accelerate other modeling efforts it would also be a good tool to have.

berkgercek:
Just bumping this PR: it contains a function for loading trial info into a pandas dataframe which others have been asking for, so I'd like to share it to develop ASAP

PR #247, opened 2020-11-17
Small fixes 

guidomeijer:
Sorry forgot two small things!


guidomeijer:
As requested I also added the possibility to only plot one hemisphere in the atlas plots

nbonacchi:
you forgot to flake :P 
$ flake8 .
./brainbox/population/population.py:311:71: E721 do not compare types, use 'isinstance()'

PR #251, opened 2020-11-19
standard trials extractors return complete data

k1o0:
I've refactored the extractor code.  

Main differences:
- Task QC Extractor calls either `FPGATrials` or the training `extract_all` function instead of individual extractor functions.
- The main training extraction function was moved out of `pipes.training_preprocessing` to avoid circular imports
- The trails extractors extract all relevant data, including trigger times, etc.
- The extractor classes may have `save_names` that are None.  If the filename is None, the data aren't saved but are still returned

The reasons for this:
- The QC extractor is now directly using the trial extractors, which will reduce QC extraction errors and avoid divergence of the extractors
- Extracting the extraneous trial data doesn't take very long and in doing it this way data aren't re-extracted/loaded more than once.

I haven't run integration tests yet.


ibllib/io/extractors/base.py
```diff
@@ -67,16 +72,19 @@ def _write_to_disk(file_path, data):
             else:
                 log.error(f"Don't know how to save {file_path.suffix} files yet")
 
-        if isinstance(self.save_names, str):
+        if self.save_names is None:
```
oliche, line 75: Here I had to distinguish the case were none of the output were requested, all, and partial

ibllib/io/extractors/bpod_trials.py
```diff
@@ -0,0 +1,55 @@
...
+    _logger.info(f"Extracting {session_path} as {extractor_type}")
+    bpod_trials = bpod_trials or rawio.load_data(session_path)
+    settings = settings or rawio.load_settings(session_path)
+    if extractor_type == 'training':
+        _logger.info('training session on ' + settings['PYBPOD_BOARD'])
+        wheel, files_wheel = training_wheel.extract_all(
+            session_path, bpod_trials=bpod_trials, settings=settings, save=save)
+        trials, files_trials = training_trials.extract_all(
+            session_path, bpod_trials=bpod_trials, settings=settings, save=save)
+    elif extractor_type == 'biased' or extractor_type == 'ephys':
```
oliche, line 36: for a bpod only ephys extraction, this should be directed to the biased trials extractor

ibllib/pipes/ephys_preprocessing.py
```diff
@@ -232,11 +232,10 @@ def _run(self):
 
         # Run the task QC
         qc = TaskQC(self.session_path, one=self.one, log=_logger)
-        qc.extractor = TaskQCExtractor(self.session_path, lazy=True, one=qc.one)
-        # Extract extra datasets required for QC
-        qc.extractor.data = dsets
-        qc.extractor.extract_data(partial=True)
-
+        qc.extractor = TaskQCExtractor(self.session_path, one=qc.one)
```
oliche, line 235: The lazy option would miss adding the `quiescence` in the dictionary and omit the `self.settings` property

ibllib/pipes/training_preprocessing.py
```diff
@@ -31,32 +29,23 @@ def _run(self):
...
-            qc.extractor.data = trials
-            qc.extractor.extract_data(partial=True)  # Extract the rest of the data
-            return qc
-
-        if get_session_extractor_type(self.session_path) == 'habituation':
+        type = get_session_extractor_type(self.session_path)
+        if type == 'habituation':
             qc = HabituationQC(self.session_path, one=self.one)
-            qc = _qc_extract(qc)
+            qc.extractor = TaskQCExtractor(self.session_path, one=self.one)
```
oliche, line 41: The lazy option would miss adding the `quiescence` in the dictionary and omit the `self.settings` property

ibllib/io/extractors/training_wheel.py
```diff
@@ -395,11 +396,12 @@ def _extract(self):
...
+        first_moves, is_final, _ = extract_first_movement_times(moves, trials, min_qt=min_qt)
+        output = (ts, pos, moves['intervals'], moves['peakAmplitude'],
+                  moves['peakVelocity_times'], first_moves, is_final)
+        return output
 
 
 def extract_all(session_path, bpod_trials=None, settings=None, save=False):
-    return Wheel(session_path=session_path).extract(
-        save=save, bpod_trials=bpod_trials, settings=settings)
+    return run_extractor_classes(Wheel, save=save, session_path=session_path,
```
oliche, line 406: The extract all method should return an ordered dict, not a list

oliche (changes requested):
Unit tests were failing as were Integration tests.
It's still a work in progress.

So far I've done the minimum necessary to get the ephys pipeline integration test to run so I can work on refactoring the ephys_fpga trials extractor.

The training pipeline integration tests and the QC tasks are still failing, partly because of sessions prior to 5.0.0 that are missing some datasets. 
One mandatory thing is to add > 5.0.0 sessions in the integration suite to test the full task qc. One optional is to ensure old sessions are still extracted. I don't think it's going to be a lot of work, so I would try and maintain compatibility for the time being.

PR #260, opened 2021-01-05
Changes to Globus Patcher 

oliche:


rossant:
LGTM
FYI I had started a thin Globus wrapper with tests in my branch https://github.com/int-brain-lab/ibllib/blob/globus/ibllib/io/globus.py the idea being to decouple the Globus code from the alyx code. You can also see the `@blocking` decorator to implement waiting. In any case it's important to test independently the different functionalities!

PR #269, opened 2021-02-02
Fix breaking of onelib.params.setup() 

juhuntenburg:
HI, a small fix for a problem I encountered setting up ONE. I am not sure about the protocol for contributing yet, so just opening the PR to get feedback. In principle I assume I should write a test to reproduce the problem, which is fixed by this PR?


juhuntenburg:
Note to myself: this should be changed in the docs too

juhuntenburg:
Added a test for the setup function which fails in the original code and passes with this fix. Also rebased

PR #271, opened 2021-02-10
Created new functions 'classify' and 'regress' and added deprication warning to 'decode'

guidomeijer:
The 'decode' function was getting out of hand, I tried to incorporate all kinds of functionality in it and ended up always having to add more functionality. So I decided to deprecate the decode function altogether. It still works but gives a deprivation warning, I'll remove it in the next PR. Instead I created the 'classify' and 'regress' functions which are very basic, users can create whatever functionality they want (sub-selecting neurons etc) themselves and only call the decoding function when doing the actual decoding. I also updated the example to reflect this.


k1o0:
In `TestTask/test_get_spike_counts_in_bins` it's still calling `_get_spike_counts_in_bins` (with underscore).  Also you need to add a unit test to increase the coverage ;)

guidomeijer:
I removed _get_spike_counts_in_bins from brainbox.task altogether and only use get_spike_counts_in_bins from brainbox.population now. I also added some unit tests.

k1o0:
Sorry the CI isn't very clear.  I'll change it so that flake8 errors are reflected in the description, instead of just `1`.  If you click on [Details](https://ibllib.hooks.internationalbrainlab.org/logs/e9e3ecabe0ff282cdac4937b77fb8a8480baa69f) and scroll to the bottom you'll see the output of flake.  `test_population` and `task` are both failing PEP8.

guidomeijer:
Sorry I keep messing up the flaking, should be good now. What is the best way to do flaking? My built-in Spyder code checker is not very good and misses a lot of these flaking errors and when I do: 
```
conda activate iblenv
flake8 /path/to/scripts
```
I get a lot of "line too long" errors which are fine because we use 99 characters instead of 79

k1o0:
Yeah it's annoying.  You have to call flake8 from within the ibllib repo folder.  When you do that it reads the .flake8 config file.  You can also call it with the config path from any folder: `flake8 /path/to/scripts --config=/path/to/ibllib/.flake8`.  There's only on line failing now.

guidomeijer:
Thanks that works much better

k1o0:
Nearly there! There's quite a few new functions in task.py that aren't covered by tests.  I also notice that `generate_pseudo_blocks` exists in both task.py and population.py.  One has documentation and the other has an extra param.  Could they be merged?

guidomeijer:
I just removed generate_pseudo_blocks from population altogether because it was only being used by the decode function which will be deprecated anyway

k1o0:
I've merged.  Please delete the branch if you're not going to use it anymore

PR #273, opened 2021-02-19
create qc field on insertion creation

mayofaulkner:
make qc and extended qc field on probe insertion creation


k1o0:
You're lucky the coverage check passed :P

PR #275, opened 2021-02-26
few utils functions to get file paths from alyx rest models

oliche:


k1o0:
I'm not sure if this is relevant but on Windows the globus_sdk still requires paths to be in the Unix path convention

k1o0:
Looks like this was opened when the test instance was maxed out.  Here are the status checks:
[Tests status badge](https://ibllib.hooks.internationalbrainlab.org/logs/records/sdsc)
[Coverage badge](https://ibllib.hooks.internationalbrainlab.org/coverage/sdsc)

oneibl/webclient.py
```diff
@@ -12,6 +12,7 @@
 
 from ibllib.misc import pprint, print_progress
 
+SDSC_ROOT_PATH = Path('/mnt/ibl')
```
k1o0, line 15: This should be a parameter, rather than being hard-coded, or better yet, it could be taken from Alyx

oneibl/webclient.py
```diff
@@ -53,6 +54,74 @@ def __iter__(self):
...
+    if repository:
+        fr = next((fr for fr in dset['file_records'] if fr['data_repository'] == repository))
+    else:
+        fr = next((fr for fr in dset['file_records'] if fr['data_url']))
+    uuid = dset['url'][-36:] if uuid else None
+    return _path_from_filerecord(fr, root_path=root_path, uuid=uuid)
+
+
+def _path_from_filerecord(fr, root_path=SDSC_ROOT_PATH, uuid=None):
+    import alf.io
```
k1o0, line 113: What's the reason for importing within the function?

oliche: those nasty brainbox circular imports

PR #282, opened 2021-03-12
Added functionality to return decoding accuracy on the training set

guidomeijer:
Just added some extra functionality to the classify function


k1o0:
Hi Guido, before we merge please could you update the tests to cover your changes?

PR #287, opened 2021-03-22
Enh/task

juhuntenburg:
Adding two new arguments to `Task.__init__()` , passed on to `Task.run()` , as well as the `Pipeline.run()` and `Pipeline.run_alyx_task()` function. 
1. `machine`: string to tag the machine a task or pipeline is run on. Defaults to None, if not None, will be logged.
2. `clobber`: boolean to indicate whether or not existing logs on alyx should be overwritten. This defaults to False, meaning that the **default behavior will be changed** by this PR. 

Happy for any comments!

### DIFF
--- a/ibllib/pipes/tasks.py
+++ b/ibllib/pipes/tasks.py
@@ -31,7 +31,7 @@ class Task(abc.ABC):
     time_out_secs = None
     version = version.ibllib()
 
-    def __init__(self, session_path, parents=None, taskid=None, one=None):
+    def __init__(self, session_path, parents=None, taskid=None, one=None, machine=None):
         self.taskid = taskid
         self.one = one
         self.session_path = session_path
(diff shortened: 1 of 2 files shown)


ibllib/pipes/tasks.py
```diff
@@ -57,6 +61,8 @@ def run(self, **kwargs):
         if use_alyx:
             self.one.alyx.rest('tasks', 'partial_update',
                                id=self.taskid, data={'status': 'Started'})
```
juhuntenburg, line 63: return tdict here do not hit database again

ibllib/pipes/tasks.py
```diff
@@ -57,6 +61,8 @@ def run(self, **kwargs):
...
+            pre_log = self.one.alyx.rest('tasks', 'list', id=self.taskid)[0]['log']
+            self.log = pre_log + '\n\n' if pre_log else ''
```
juhuntenburg, line 65: Make header for new logs unique pattern

juhuntenburg:
Fixed the things we discussed today, once the checks pass I think it is ready to merge?

PR #289, opened 2021-03-23
Return GPIO as nx4 bool array

k1o0:
Following Nicco's revelation that the pin state values encode the 4 pin states I've changed the loader to return a boolean array with the shape (n_frames, n_pins), where n_pins is 4.  

I don't consider bit shifting as 'extraction' any more than typecasting.  To read the binary file in any other form, simply load it directly (e.g. with a numpy loader).  

The as_dict argument (default=False) now returns a tuple of 4 dicts, one per pin, or a None for the pins that were inactive.

### DIFF
--- a/brainbox/tests/test_core.py
+++ b/brainbox/tests/test_core.py
@@ -41,6 +41,8 @@ class TestBunch(unittest.TestCase):
             abunch.save(npz_filec, compress=True)
             another_bunch = core.Bunch.load(npz_filec)
             [self.assertTrue(np.all(abunch[k]) == np.all(another_bunch[k])) for k in abunch]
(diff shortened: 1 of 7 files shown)


nbonacchi:
Except or the tests failing, the loader looks fine!
Except for debugging purposes I don't think we'll ever want to load the un-bitshifted value.
I might probably add this to the docstring something like: 
...
The raw output file binary format is a double precision float. The integer part represents the state of the 4 gpio pin states of the camera. Uses a bitshift of 27, 28, 29, and 30 for the gpio state of port 3, 2, 1, and 0 respectively. To load the outputted integer use `np.fromfile(file_name, dtype=np.float64).astype(int)`
...

PR #290, opened 2021-03-24
Brainbox plots

mayofaulkner:


k1o0:
I merged the brainbox structuring into this branch so it's up-to-date with develop.  I updated the imports and noticed that there are untested, potentially broken functions in plot.py.  Also now that we have three plot modules (plot.py, base_plot.py and plot_ephys.py) maybe it's worth moving them back into a plot package.

Also I'm sure my refactor has broken some of the examples 🙈 Maybe we could go over it next week because even before my refactoring, building the docs fails on the CI.

PR #293, opened 2021-03-30
Orphaned audio TTLs

k1o0:
Fixed index error when removing orphaned audio TTLs in the grooming function.  I ran this on about 25 of the 34 error'd sessions and they all extracted.  

Also the old way of extracting timestamps is used as a fallback if any of the assertions fail.  Besides this and the GPIO loader issue I haven't seen any other errors that aren't genuine problems with the hardware.  The fallback will make any future failures hard to detect - you'll have to rely on the QC - but it means the video will be registered in most cases.


k1o0:
Camera extraction tests were ignored -> renamed the test file so it starts with 'test_' and re-triggered the tests.

PR #294, opened 2021-03-30
Ephys trials

oliche:
Refactored the trials extraction to extract for opto laser stimulation datasets.

The underlying change is trivial, the real change is how to accommodate for new task types and flexible extractors.
So far the solution has been to associate a set of extractor classes to a task type.


k1o0:
I'm not angry, just disappointed.

PR #296, opened 2021-03-31
GitHub actions

nbonacchi:


k1o0:
Please don't push anything to this branch until the test instance has been upgraded: it keeps becoming unresponsive.

nbonacchi:
It was responding yesterday before I wrapped the day and I haven't done any commits yet today. I think this branch might not be what caused this were you running tests?

k1o0:
Yeah you're probably right then.  I suppose anyone running tests locally can put pressure on it, plus dumping the db at midnight might cause issues, especially if it's also serving data via the CI tunnel...

PR #297, opened 2021-04-01
Brainbox plots

mayofaulkner:
refactored examples to work with new brainbox import structure


k1o0:
Hmm I don't know why the coverage check is still pending.  I can see from the logs that it passed and posted a 'success' status to Github.  Just ignore and merge when you're ready.

PR #298, opened 2021-04-01
Allowing cap as input, adding quiet flag

juhuntenburg:


ibllib/io/video.py
```diff
@@ -63,20 +64,18 @@ def get_video_frames_preload(video_path, frame_numbers=None, mask=Ellipsis, as_l
         frame_images = [None] * len(frame_numbers)
     else:
         ret, frame = cap.read()
-        frame_images = np.empty((len(frame_numbers), *func(frame[mask or ...]).shape),
-                                np.uint8)
+        frame_images = np.zeros((len(frame_numbers), *func(frame[mask or ...]).shape), np.uint8)
+    print(f"Loading {len(frame_numbers)} frames")
     for ii, i in enumerate(frame_numbers):
-        sys.stdout.write(f'\rloading frame {ii}/{len(frame_numbers)}')
```
k1o0, line 69: Please restore the frame loader.  What were @oliche's reasons for removing it?  I'm fine with a quiet flag that is False by default.

ibllib/io/video.py
```diff
@@ -31,7 +30,7 @@ def get_video_frame(video_path, frame_number):
 
 
 def get_video_frames_preload(video_path, frame_numbers=None, mask=Ellipsis, as_list=False,
-                             func=lambda x: x,):
+                             func=lambda x: x, cap=None):
```
k1o0, line 33: It doesn't make sense to provide both a video path and and a VideoCapture class, especially as the video_path is a positional argument.  Instead perhaps the first input can be one of (cv2.VideoCapture, VideoStreamer, str, Path)?

k1o0 (changes requested):

ibllib/io/video.py
juhuntenburg: Ah yes good point. I assume the parameter name cannot be changed though, right? (As in I cannot call it video_path_or_cap to be more explicit)?
I will adapt the code and try to make it clear in the docstring then

ibllib/io/video.py
juhuntenburg: The argument was that it slows things down to print to the command line

ibllib/io/video.py
```diff
@@ -31,25 +31,31 @@ def get_video_frame(video_path, frame_number):
...
     Default dimensions are (n, w, h, 3) where n = len(frame_numbers)
 
     Example - Load first 1000 frames, keeping only the first colour channel:
         frames = get_video_frames_preload(video_path, range(1000), mask=np.s_[:, :, 0])
     """
-    is_url = isinstance(video_path, str) and video_path.startswith('http')
-    cap = VideoStreamer(video_path).cap if is_url else cv2.VideoCapture(str(video_path))
+    is_cap = isinstance(video_path, (cv2.VideoCapture, VideoStreamer))
+    if is_cap:
+        cap = video_path
```
k1o0, line 55: Could we change the input name to 'vid'?  It's more accurate.  I'm not sure this will work... if video_path is an instance of VideoStreamer you need to get the cap handle from the cap property, something like:
```
if isinstance(video_path, cv2.VideoCapture):
    cap = video_path
elif  isinstance(video_path, VideoStreamer):
    cap = video_path.cap
else:
        is_url = isinstance(video_path, str) and video_path.startswith('http')
        cap = VideoStreamer(video_path).cap if is_url else cv2.VideoCapture(str(video_path))
```

ibllib/io/video.py
k1o0: Printing to the stdout is _extremely_ fast.  In fact it's even faster than a print statement:
```
import time
import sys
n = 10000
times = []
print('========== Starting print test ==========')
for _ in range(3):
    start_time = time.time()
    for i in range(n):
        print(f'\rloading frame {i}/{n}')
    times.append(time.time() - start_time)
print(times)
print(sum(times) / 3)
times = []
print('========== Starting stdout test ==========')
for _ in range(3):
    start_time = time.time()
    for i in range(n):
        sys.stdout.write(f'\rloading frame {i}/{n}')
        sys.stdout.flush()
    times.append(time.time() - start_time)
print(times)
print(sum(times) / 3)
"""
========== Starting print test ==========
[5.338913440704346, 1.0393381118774414, 1.0872581005096436]
2.4885032176971436
========== Starting stdout test ==========
loading frame 9999/10000[0.47832822799682617, 0.4129786491394043, 0.4161257743835449]
0.4358108838399251"""
```

Also it's extremely useful when you're sampling non-contiguous frames, especially over http.  That can take many minutes so it's good to know where you are.  The camera QC samples 100 frames by default and that takes over 10 minutes.

k1o0: I'm pretty confident that any performance improvement you see is a result of not having to re-open the video file.  I don't know how many times you call this function but that quickly adds up, as does getting the video length.  (FYI the VideoStreamer object stores it as an attribute).  When I wrote the function I intended for it to be called once per video so I didn't optimize it for other use cases.

ibllib/io/video.py
juhuntenburg: > Could we change the input name to 'vid'? 

Ah you don't mind changing the argument name? Sure I can do that. I was originally thinking "video_path_or_cap", but "vid" is simpler :)

> if video_path is an instance of VideoStreamer you need to get the cap handle from the cap property

How about just allowing a cv2.VideoCapture instance? For example in the motion energy function I can use this same line
`cap = VideoStreamer(video_path).cap if is_url else cv2.VideoCapture(str(video_path))`
and then just input the result to preload_video_frames

ibllib/io/video.py
juhuntenburg: Should all be solved with the quiet flag, so one can pick their preference

ibllib/io/video.py
k1o0: You can rename the positional input variables to what ever you like, it won't break anyone's code.  I suggested 'vid' because it encompasses any representation of a video, including the VideoStreamer object, but it's up to you - it's not that important.

juhuntenburg:
Not clear about this error: 
```
is_cap = isinstance(vid, cv2.VideoCapture)
TypeError: isinstance() arg 2 must be a type or tuple of types
```
This works fine for me locally

ibllib/io/video.py
k1o0: > How about just allowing a cv2.VideoCapture instance? 
Only your function provides a cv2.VideoCapture object because it's the only one that calls this function many times over.  If you narrow the input types then you have refactor everyone else's code and we'll end up with more lines overall because everyone will have to open their videos within the calling function.  The if-elif-else statement in the comment above should cover all current input types, including a VideoCapture handle.

juhuntenburg: No I think that was a misunderstanding. I wasn't suggesting to remove the option of inputting a Path of course. Just the option of inputting an VideoStreamer instance. My suggestion (as implemented in the last version) is to allow either a local path, a URL, or a VideoCapture instance as input. If one wants to create a VideoStreamer instance outside of the function they can create it and then pass VideoStreamer.cap.  That shouldn't break anyones code as it only adds functionality

juhuntenburg:
> Not clear about this error:
> 
> ```
> is_cap = isinstance(vid, cv2.VideoCapture)
> TypeError: isinstance() arg 2 must be a type or tuple of types
> ```
> 
> This works fine for me locally

I think it is because the cv2.VideoCapture class is overwritten by the mock in those three tests (see [here](https://github.com/int-brain-lab/iblscripts/blob/develop/ci/tests/test_camera_extraction.py#L500))

k1o0:
> 
> 
> > Not clear about this error:
> > ```
> > is_cap = isinstance(vid, cv2.VideoCapture)
> > TypeError: isinstance() arg 2 must be a type or tuple of types
> > ```
> > 
> > 
> > This works fine for me locally
> 
> I think it is because the cv2.VideoCapture class is overwritten by the mock in those three tests (see [here](https://github.com/int-brain-lab/iblscripts/blob/develop/ci/tests/test_camera_extraction.py#L500))

Mocking with `spec` might solve this: https://stackoverflow.com/questions/11282401/can-i-fake-mock-the-type-of-my-mock-objects-in-python-unittests/11283173#11283173

PR #310, opened 2021-04-21
Session qc reason

GaelleChapuis:
main function to be called if a user wants to label a session as CRITICAL


k1o0:
I'm closing this PR.  Each time you push a commit you trigger the tests and the queue is getting long.  Also you will have to re-open it as a PR to develop anyway.  Please only open the PR when you're sure everything is finished and tested.

PR #316, opened 2021-04-28
Set save name to None for non-essential failed datasets

k1o0:
Fix for https://github.com/int-brain-lab/iblenv/issues/232


nbonacchi:
Looks good, I have been meaning to fix that! thanks.
I've ran into the issue of extractors saving empty files I though that returning None as the data would prevent the saving, but the behavior of base class is to save all file names even if data is None. So if we want to extract what we can even if something fails adding a last check to set the names as None and not the data as None is how to do it.
(This is more of a comment for me :smile:)

PR #319, opened 2021-05-03
Large overhaul of GLM codebase

berkgercek:
The GLM code base was getting very, very ugly. I had implemented a linear, rather than nonlinear, model but subclassed the non-linear model rather than a proper refactoring.

This PR will basically change the entire API for the brainbox.modeling module. Shouldn't affect anything else.

In brief:

- Design matrix generation and fitting of neural data are performed by two separate classes, design_matrix.DesignMatrix and either linear.LinearGLM or poisson.PoissonGLM
- Both the linear model and poisson model above inherit a common class called NeuralModel which has most of the methods necessary for treating data
- Some utility functions have been moved to brainbox.modeling.utils, in particular a class called SequentialSelector which allows you to run stepwise regressor selection on the data. This used to be a nasty chunk of code in the GLM object but I rewrote from scratch and based it on sklearn.feature_selection.SequentialFeatureSelector, which doesn't support feature groups (which is what I need) or intermediate-model scoring.

I also removed the tests for the modeling submodule for now, because I need to give some serious thought as to what to put there versus in the integration tests for ibllib.


berkgercek:
@k1o0 @oliche  I'm not sure why the build is failing here, but it doesn't seem to be coming from any change I've made... The failure comes from oneibl tests

Suggestions?

berkgercek:
Ah I just read your previous message @k1o0 , so I understand why the github checks are failing. What happened to the UCL ones? Are those only for changes to Master?

k1o0:
The CI checks are [currently running](https://ibllib.hooks.internationalbrainlab.org/log/a5dac6d36908b8752bb78e30f064021d9bb2490f) for the second-to-last commit (`a5dac6d`).  It skipped the last commit because it was just a change to the readme.  I thought GitHub would display the last update for each status check but apparently not...

k1o0:
The UCL CI checks pass.  Is there currently any test coverage of the modelling module?

berkgercek:
No, I removed the coverage for now because the test was badly outdated and I need to take some time to split apart my tests into integration tests (going into iblscripts) and unit tests. They're too bulky to be used as unit tests ATM.

berkgercek:
@k1o0 Do I need to write tests ASAP to keep the code cov % static or can I put them off until later?

k1o0:
> @k1o0 Do I need to write tests ASAP to keep the code cov % static or can I put them off until later?

I'll merge for now but there are plenty of functions that could have simple unit tests added, such as the utils or the score method.

PR #330, opened 2021-05-20
New raw_data_loader func and extractor integration for *Camera.frameData.bin files

nbonacchi:


k1o0:
We should review the dataset type first.  For one the name should be in Haskell case, i.e. 'leftCamera.frameData.bin'.  The extractors and CI tests will need updating before any sessions are acquired with this dataset.

nbonacchi:
Sure that's just there for when we get the release candate data to be able to look at it...
Now that I look at it this should be a draft PR not a full PR...

ibllib/io/raw_data_loaders.py
```diff
@@ -103,16 +103,75 @@ def load_data(session_path: Union[str, Path], time='absolute'):
...
+            }
+    """
+    camera = assert_valid_label(camera)
+    fpath = Path(session_path).joinpath("raw_video_data", f"_iblrig_{camera}Camera.frameData.bin")
+    assert fpath.exists(), f"{fpath}\nFile not Found: Could not find bin file for cam <{camera}>"
+    rdata = np.fromfile(fpath, dtype=np.float64)
+    assert len(rdata) % 4 == 0, "Dimension mismatch: bin file length is not mod 4"
+    rows = int(len(rdata) / 4)
+    data = np.reshape(rdata.astype(np.int64), (rows, 4))
+    df_dict = {
```
k1o0, line 140: Could use fromkeys for fewer lines :)

ibllib/io/raw_data_loaders.py
```diff
@@ -103,16 +103,75 @@ def load_data(session_path: Union[str, Path], time='absolute'):
...
+        "Timestamp": None,
+        "embeddedTimeStamp": None,
+        "embeddedFrameCounter": None,
+        "embeddedGPIOPinState": None
+    }
+    df = pd.DataFrame(data, columns=df_dict.keys())
```
k1o0, line 146: Could add a return here as there's no need to parse the df if raw is True

k1o0 (approved):

nbonacchi:
OK integrated your suggestions, ready for merge!

PR #336, opened 2021-06-09
ONE2

k1o0:
- ONE is now a dependency.
- Functions common to ibllib and ONE are now in a third dependency, iblutil.
- All internal functions use the new ONE.
- Importing oneibl or alf raises a deprecation warning.
- A couple of brainbox functions check for the old One instance and call the deprecated package if required.
- oneibl.registration is still imported.  This should be generalized and moved to ONE.
- REST GET requests are now cached.  Instantiate with cache_rest=None to deactivate; or use no_cache=True in rest method.
- The test database is now configurable with an env var.  Test URL imported from a single location.


k1o0:
I'm merging for now but will keep the branch for updating examples

PR #338, opened 2021-06-16
atlas_download

mayofaulkner:
- Changed the AllenAtlas function so that by default it will download from the Allen institue
- Added try except which will setup one params to default public ones if the params file does not already exist (useful for external users using the alignment gui and iblviewer!)


k1o0:
The params have changed a bit in ONE2 but I'll copy your changes to the ONE2 branch

k1o0:
I applied these changes to the ONE2 branch, which is now passing.  I'm going to close this one instead of merging to avoid conflicts down the line.

PR #340, opened 2021-06-23
ONE2 in examples

k1o0:
I've updated most of the example scripts with the new ONE.  I gave up on some of the older ones.


oneibl/registration.py
```diff
@@ -59,7 +59,7 @@ def register_dataset(file_list, one=None, created_by=None, repository=None, serv
     :return:
     """
     if created_by is None:
-        created_by = one.alyx.user
+        created_by = one._par.ALYX_LOGIN
```
k1o0, line 62: @oliche this isn't going to work with the new ONE.  Did you change it back because you ran it with a oneibl.One instance?  If so for your legacy code call register_file with the created_by argument.

PR #342, opened 2021-06-29
DLC stuff

nbonacchi:


k1o0:
Let's discourage the use of alyx REST queries when it's not required. This query can be done with ONE.search.  Maybe it would be better to return sessions based on the QC anyway.  All valid video sessions should have QC by now, and the timestamps QC check is more accurate that the mere presence of a GPIO file.  There are some sessions where the GPIO or raw timestamps files aren't good so the extractor falls back on the old extraction method.

nbonacchi:
I opened this for you at the time @mschart is this dead or did you integrate it in some other way already? Can I close?

PR #344, opened 2021-07-06
One2 docs

mayofaulkner:
- updated docs to work with ONE2
- made ONE2 version of brainbox.io.load_spike_sorting the same as previous behaviour


k1o0:
Looks like you edited the deprecated version of brainbox.io.one! The file in brainbox.io.deprecated.one is the old version that still works with oneibl.one.  If a user passes an old instance of One to the function, then a warning is thrown and the deprecated module is called.  There was a typo that I fixed yesterday (on the ONE2 branch) that meant that the new One called the deprecated function (I missed a 'not' in the if statement): https://github.com/int-brain-lab/ibllib/pull/340/commits/2b9b78a3f6a10ce4cf493d26c58a163cf2081f98.  We should keep the old brainbox function unchanged while users migrate to the new ONE.  

Could you revert those changes and push the other changes to the ONE2 branch?

mayofaulkner:
Okay, have reverted this and merged ONE2 into my branch. Will make a PR to ONE2 instead of develop

PR #345, opened 2021-07-06
One2 docs

mayofaulkner:


brainbox/io/one.py
```diff
@@ -359,11 +381,11 @@ def load_passive_rfmap(eid, one=None):
...
         one = one or ONE()
         dtypes = {
-            '_iblrig_RFMapStim.raw',
-            '_ibl_passiveRFM.times',
+            '_iblrig_RFMapStim.raw.bin',
+            '_ibl_passiveRFM.times.npy',
         }
 
-        _ = one.load_dataset(eid, dataset_types=dtypes, download_only=True)
+        _ = one.load_datasets(eid, datasets=dtypes, download_only=True)
```
k1o0, line 388: This function (`load_passive_rfmap`) should be updated for the new ONE.  Now it doesn't matter whether eid is a UUID or a path.  Instead of calling load_datasets then calling alfio methods it should simply call `one.load_object(eid, 'passiveRFM')` and `one.load_dataset(eid, '_iblrig_RFMapStim.raw.bin')`  This would reduce the number of lines by a lot.

brainbox/examples/docs_explore_passive.py
```diff
@@ -12,14 +12,14 @@
 import brainbox.io.one as bbone
 import brainbox.task.passive as passive
 
-eid = 'c7248e09-8c0d-40f2-9eb4-700a8973d8c8'
-probe = 'probe01'
-one = ONE()
+eid = '4ecb5d24-f5cc-402c-be28-9d0f7cb14b3a'
+probe = 'probe00'
+one = ONE(base_url='https://openalyx.internationalbrainlab.org')
```
k1o0, line 17: It might be worth adding the silent flag to all these, that way the user will never be prompted to set up open alyx

brainbox/examples/docs_load_spike_sorting.py
```diff
@@ -37,14 +37,7 @@
...
-
-##################################################################################################
-# Example 4:
-# By default, load_spike_sorting_with_channel will look for available data on your
-# local computer and if all the datasets are found it will not check for consistency with files
-# stored on flatiron (to speed up loading process). To make sure that data is always synced to
-# flatiron we can set a force flag to True
-spikes, clusters, channels = bbone.load_spike_sorting_with_channel(eid, one=one, probe='probe00')
+spikes, clusters, channels = bbone.load_spike_sorting_with_channel(eid, one=one, probe='probe01',
+                                                                   dataset_types=['clusters.peakToTrough'])
```
k1o0, line 41: I was a bit cavalier in refactoring `load_spike_sorting` and removed the dataset_types kwarg. If you think it's worth keeping perhaps rename it to `datasets` instead.  At the moment it will download the full spikes object.  For now you could use `|` in attribute to load selected attributes (I still haven't updated it to support lists):
```
default_attr = [
    'times',
    'amplitudes'
]
spikes = one.load_object(eid, 'spikes', attribute='|'.join(default_attr), collection=f'alf/{probe_label}')
```

k1o0 (commented):
If you like I can have a crack at refactoring `load_passive_rfmap`?

brainbox/io/one.py
mayofaulkner: Okay, good point, I can refactor this

brainbox/examples/docs_explore_passive.py
mayofaulkner: Good idea!

brainbox/examples/docs_load_spike_sorting.py
mayofaulkner: This one, I have already changed so it maintains the same behaviour as the old code. I could be inclined to keep the argument as dataset_types just to causes as little disruption to user code as possible. But also happy to change if you have strong opinions otherwise :)

PR #348, opened 2021-07-13
work with 3A and 3B probes

mayofaulkner:


ibllib/qc/task_extractors.py
```diff
@@ -64,31 +64,42 @@ def _ensure_required_data(self):
...
+                            'ephysData.raw.wiring'])
+
+            if len(self.one.list_datasets(eid, collection='raw_ephys_data')) == 0:
+                collections += ('raw_ephys_data/probe00', 'raw_ephys_data/probe01')
+            else:
+                collections += ('raw_ephys_data',)
+
+        files = []
+        for dstype in dstypes:
+            dataset = self.one.datasets_from_type(eid, dstype, full=True)
```
k1o0, line 98: Tomorrow this will be `dataset = self.one.type2datasets(eid, dstype, details=True)`.  You could remove the for loop as now it permits a list of dataset types (should be quicker)

PR #358, opened 2021-08-03
add opto extraction for widefieldChoiceWorld

cskrasniak:
can I still use the extractor type "ephys_biased_opto" (which stipulates that the biased blocks are not pre-defined) even if I am using the predefined blocks without error? or will the added extraction of block and contrast cause errors with the predefined blocks? If you think this will be a problem, please reject pull request.


oliche:
I'll put the change in my pykilosort branch so I do a single release ! 
Thanks !

PR #366, opened 2021-09-09
Update README.md to say this software is deprecated, and link the new ONE2 version

kdharris101:


k1o0:
This package has now been removed so no need to update the README

PR #368, opened 2021-09-13
use timing info of spacer to filter out erroneously detected spacers

mayofaulkner:


nbonacchi:
Oh wow!
That's nice!!! Great idea @mayofaulkner !!
I was thinking of timing in the sense that the periods they define are always ~the same, i.e. 3 chunks of 5 min each, but testing for monotonic increase & decrease of the spacer itself is better!!

Did you write an integration test for this? You could add one of the 4 spacers sessions (required files) to the integration data and run this to find 3 spacers!!! Or if the DL is fast enough use something like a `download_required_datasets(eid, task_name)` and/or a `required_datasets_exist(eid, task_name)` funcs to do that on demand...
What do you think?

PR #372, opened 2021-09-17
Handle empty timestamps in video compression and qc

juhuntenburg:


juhuntenburg:
Once this is merged and updated on cortexlab, reset EphysVideoCompression task (id: `'65edaada-49ef-44d1-a95e-212794b87bbc')` of session cortexlab/Subjects/KS055/2021-05-03/001 (id: `'872ce8ff-9fb3-485c-be00-bc5479e0095b'`)

k1o0:
Looks like you need to deal with a None error.  Merge develop into this branch to get rid of the other two errors.

k1o0:
New test failure:
```
======================================================================
FAIL: test_incomplete_session (test_camera_extraction.TestCameraQC)
  File "/home/experiment/Documents/github/iblscripts/ci/tests/test_camera_extraction.py", line 544, in test_incomplete_session
    self.assertEqual('FAIL', outcome)
AssertionError: 'FAIL' != 'NOT_SET'
- FAIL
+ NOT_SET
```

juhuntenburg:
So the issue is that with the change I made the "incomplete session" will not run at all, since there are no timestamps. I would need to set the expected outcome to 'NOT SET' and remove all the expected values for extended qc. Which makes me doubt if that is the way to go? Alternatively maybe it is worth adding camera.times to the incomplete dataset. What do you think @k1o0 ?

juhuntenburg:
I fixed this test, how do I re-trigger the UCL tests? Also, can you double check if this is ok @k1o0 ? I had to remove the entire extended QC output because the lack of timestamps in the incomplete session now means that the checks won't run at all.

k1o0:
As long as the coverage doesn't go down, it's fine

juhuntenburg:
Great, all passed. I'll merge this

k1o0:
Done! Thanks for this

PR #380, opened 2021-10-04
Rms ephysqc

juhuntenburg:
Adding EphysQC class and running it in RawEphysQC task


juhuntenburg:
I think the decrease in coverage is because I am not actually running the qc functions. It is hard to unittest this, but I could run them in the integration test (by just removing the existing outputs from the integration data). What do you think?

PR #381, opened 2021-10-05
AllenAtlas always returns coordinates in um

k1o0:
Fix for issue https://github.com/int-brain-lab/iblenv/issues/261. 

The problem appears to be that the BrainAtlas class operates in meters, while its subclass, AllenAtlas, operates in micrometers.  Additionally some BrainAtlas methods return values in micrometers (i.e. extent and plot_tilted_slice).  Other classes such as Insertion expect inputs to be micrometers.  I added an assert in Insertion._get_surface_intersection as it assumes the atlas is the Allen one (assumes units in um; uses the subclass property res_um).

Perhaps the output units should be a class property (similar to the scaling factor) and the internal units should be consistent.

### DIFF
--- a/ibllib/atlas/atlas.py
+++ b/ibllib/atlas/atlas.py
@@ -12,8 +12,8 @@ from one.webclient import http_download_file
 import one.params
 
 _logger = logging.getLogger('ibllib')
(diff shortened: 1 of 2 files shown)


oliche:
This has potentially quite big repercussions for the histology pipeline, so I would check with Mayo.
The initial agreement was to use exclusively SI for internal units.

PR #385, opened 2021-10-07
Single source ibllib version

nbonacchi:
Added:
`__version__ = "2.0.10"`
to **ibllib/__init__.py**

Added get_version(relative_path) func to **setup.py** used in version declaration of setup()

Following syntax all works:
`from ibllib import __version__` or 
`from ibllib import __version__ as ver` or 
`import ibllib` and just call `ibllib.__version__`

Added reference to release notes and reordered so latest release is on top (reverse chronological order)

Autoformatted using black


nbonacchi:
The setup.py also runs on publish to pypi, isn't that enough?

PR #410, opened 2021-11-09
Update held dependent tasks on complete

k1o0:
Upon completing a task, checks whether the parent tasks of all held dependent tasks are all now complete.  If so, changes dependent task status from 'Held' to 'Waiting'.


oliche (approved):
Looks good to me.
It's unclear to me that we shouldn't also set the held status to dep tasks to waiting if the current one is incomplete. But I'm unsure and it's a detail.

Good to merge.

PR #424, opened 2021-12-07
Bb impostortargets

bensonbrandon:
adding function that creates impostor targets, used in decoding to generate the null distribution


bensonbrandon:
I don't know if the failures have to do with my code or not.  It looks like I flaked properly and my tests should pass.

PR #434, opened 2022-01-17
numpy/numba versioning incompatibility

micheleangelofabbri:
Temp resolution to numpy/numba versioning incompatibility


oliche:
Yes it's fine (at least on ubuntu).
I suggest to merge this into dev instead !

PR #466, opened 2022-04-19
Record IDs

k1o0:
Not much changed, but if a user passes in an ONE instance a new instance is not created.  I left the alternative behaviour but it's perhaps misleading to create a new cache because users are likely to expect a call to `ONE()` within init.  If a user passes a session path without an ONE instance any loaded datasets will not be recorded. 

What was the reason for building a new cache again? With eid as the main datasets index loading data should be fast enough, even with a huge cache, and if a user wants to search data in a scratch folder they can do that by instantiating One with setup and a base dir: 
`one=One.setup(base_dir='path/to/scratch')`.

### DIFF
--- a/brainbox/io/one.py
+++ b/brainbox/io/one.py
@@ -10,6 +10,7 @@ from scipy.interpolate import interp1d
 
 from one.api import ONE, One
 import one.alf.io as alfio
+from one.alf.files import get_alf_path
 from one.alf import cache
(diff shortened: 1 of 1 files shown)


oliche:
The reason for the cache was to avoid using one altogether user side. Typically to load a portable folder, I'll let you know if I have issues with this.

PR #467, opened 2022-04-25
Changes for MTNN Analysis

hyundonglee1015:
Hi Berk,

I modified your code for mtnn analysis. Here are the main changes:
In load_trials_df function, I added additional arguments, "align_event" and "keeptrials", which determine the align event from which t_before and t_after are subtracted/added and allow me to set keeptrials manually.
In SequentialSelector, I added additional arguments to the init function, allowing me to manually set train and test trials.
In _get_best_new_feature, after training the GLM on test trials, it seems it's computing the scores on the training set again. I wonder if the scores have to be computed on the test trials.
I also made it return all the scores computed and store it in self.all_scores.

Please let me know what you think! 
Thanks.

- Hyun

### DIFF
--- a/brainbox/modeling/utils.py
+++ b/brainbox/modeling/utils.py
@@ -38,7 +38,7 @@ def neglog(weights, x, y):
 
 
 class SequentialSelector:
-    def __init__(self, model, n_features_to_select=None, direction='forward', scoring=None):
+    def __init__(self, model, traininds, testinds, n_features_to_select=None, direction='forward', scoring=None):
         """
         Sequential feature selection for neural models
 
@@ -67,8 +67,8 @@ class SequentialSelector:
         self.scoring = scoring
         self.delta_scores = pd.DataFrame(index=self.model.clu_ids)
         self.trlabels = self.design.trlabels
-        self.train = np.isin(self.trlabels, self.model.traininds).flatten()
-        self.test = ~self.train
+        self.train = np.isin(self.trlabels, traininds).flatten() #np.isin(self.trlabels, self.model.traininds).flatten()
+        self.test = np.isin(self.trlabels, testinds).flatten() #~self.train
         self.features = np.array(list(self.design.covar.keys()))
 
     def fit(self, progress=False):
@@ -98,7 +98,7 @@ class SequentialSelector:
             masks_set = maskdf.groupby(self.features.tolist()).groups
             for current_mask in tqdm(masks_set, desc='feature subset', leave=False):
                 cells = masks_set[current_mask]
-                new_feature_idx, nf_score = self._get_best_new_feature(current_mask, cells)
+                new_feature_idx, nf_score, all_scores = self._get_best_new_feature(current_mask, cells)
                 for cell in cells:
                     maskdf.at[cell, self.features[new_feature_idx.loc[cell]]] = True
                     seqdf.loc[cell, i] = self.features[new_feature_idx.loc[cell]]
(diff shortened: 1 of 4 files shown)


berkgercek:
Oh no I wish I had seen this sooner! The changes to load_trials_df are a good idea and very welcome. 

I've moved the main code base for the GLMs to [my personal page](https://github.com/berkgercek/neurencoding), but I have added you as a collaborator. The test/train issue in SFS is actually something that is on the issues page for that repo so the changes are super welcome! I'll work on integrating these changes into the repo today

berkgercek:
Okay so I reviewed your changes and they are solid except for the change to SFS.

You write `all_scores` as a dataframe of all of the scores, but it is [modified with every loop ](https://github.com/int-brain-lab/ibllib/blob/77046f801f9685667e21df2a2028a207b677edf2/brainbox/modeling/utils.py#L103)in the inner portion of `.fit()`, as a result you will only have the scores for regressors in the last iteration of sequential selection. 

What exactly do you want the behavior of the `.all_scores` attribute to be? Would you like it to store every intermediate model score in addition to the argmax which improves the model the most?

brainbox/modeling/utils.py
```diff
@@ -98,20 +100,22 @@ def fit(self, progress=False):
             masks_set = maskdf.groupby(self.features.tolist()).groups
             for current_mask in tqdm(masks_set, desc='feature subset', leave=False):
                 cells = masks_set[current_mask]
-                new_feature_idx, nf_score = self._get_best_new_feature(current_mask, cells)
+                new_feature_idx, nf_score, all_scores = self._get_best_new_feature(current_mask, cells)
```
berkgercek, line 103: I mentioned this in a message on the PR, but here you will be redefining the `all_scores` every loop, so you will only store a df with the final set of candidates for each unit. The rest of the values will be NaN. Is that what you wanted? Or would you prefer something like a stack of DFs for each step containing all of the intermediate scores?

brainbox/modeling/utils.py
```diff
@@ -121,9 +125,10 @@ def _get_best_new_feature(self, mask, cells):
             fitfeatures = self.features[candidate_mask]
             feat_idx = np.hstack([self.design.covar[feat]['dmcol_idx'] for feat in fitfeatures])
             mdm = self.design[np.ix_(self.train, feat_idx)]
+            mdm_test = self.design[np.ix_(self.test, feat_idx)]
             coefs, intercepts = self.model._fit(mdm, my, cells=cells)
             for i, cell in enumerate(cells):
                 scores.at[cell, feature_idx] = self.model._scorer(coefs.loc[cell],
                                                                   intercepts.loc[cell],
-                                                                  mdm, my[:, i])
-        return scores.idxmax(axis=1), scores.max(axis=1)
```
berkgercek, line 129: The reason I've hesitated in changing this so far is because technically if you do this you will have a test/train leak. [See this issue on the new repo](https://github.com/berkgercek/neurencoding/issues/4), but essentially you're informing your results with the test set which should be strictly for evaluation.

The proper thing to do would be to have a second line here where the model is evaluated yet again on the test set, and the method returns both the train and test scores. It's high up on my to-do list, likely this week

berkgercek (commented):
Overall looks good! The SFS class is tricky and the stuff I've marked should clarify why

brainbox/modeling/utils.py
hyundonglee1015: Thank you Berk! The changes you've made look great :). "full_scores" is exactly what I want, and I agree with your train/test evaluation issue.

Would it be possible to have these changes in the "develop" branch of ibllib?

berkgercek: It would be preferable if you switch your code over to use the neurencoding package; the interface hasn't changed at all (still uses the DesignMatrix and LinearGLM/PoissonGLM classes), you just need to ctrl F -> replace all of the instances of brainbox.modeling. I'm supporting that one and it has some bugfixes, and should be easy to install with conda develop.

berkgercek: Also I fixed the test-train issue in the neurencoding repo! The sfs object will now have a `.test_scores_` and `.train_scores_` method, as well as a flag `full_scores` to choose whether you save all the intermediate steps too

berkgercek: No, they just prevented some of the helper functions (`.add_boxcar` and `.add_covariate_raw`) from working at all. If you didn't run into them you're fine I think

berkgercek: Hey Hyun, let's move the discussion to the [issue here on the other repo](https://github.com/berkgercek/neurencoding/issues/4). I'm not sure which file you're talking about, in 197 and 198 of the new repo on SequentialSelector there is .loc[cell, i]?

PR #475, opened 2022-05-19
Rsync file transfer

micheleangelofabbri:


micheleangelofabbri:
I'm not entirely sure why we got the following error on continuous-integration/UCL:
```
======================================================================
FAIL: test_training (test_camera_extraction.TestCameraPipeline)
  File "/home/experiment/Documents/github/iblscripts/ci/tests/test_camera_extraction.py", line 679, in test_training
    self.assertEqual(job.status, 0)
AssertionError: -1 != 0
```

This aside, I am pretty happy with the current state of the rsync copy function. If I could get a second set of eyes to review, it would be really appreciated.

k1o0:
I'd get rid of any calls to `exit`: instead raise an exception.  Also if you're using rsync I would assume that we no longer need to manually keep Windows from going to sleep?

micheleangelofabbri:
> Also if you're using rsync I would assume that we no longer need to manually keep Windows from going to sleep?

I am not certain about Windows sleep. I typically test on VM's and they have somewhat unique power configurations.

PR #483, opened 2022-06-16
Rsync file transfer

micheleangelofabbri:


k1o0 (commented):
In the process of adding the widefield transfer script I abstracted out the logging to file and the parameter creation.  I noticed that even if the transfers fail the script will still delete the transfer flags and create the raw session flag! I've changed this behaviour so it's worth updating the test rigs at CCU.

micheleangelofabbri:
> 

Well that bug certainly went under the radar. Thanks for catching that. I'll update the CCU rig as soon as possible.

micheleangelofabbri:
My current understanding is that the widefield2 branch has taken over much of the functionality that this branch/PR was looking to accomplish. Converting to draft before simply closing (without merge) if this understanding is correct.

PR #484, opened 2022-06-22
Handling index out of volume

juhuntenburg:


juhuntenburg:
@oliche yours passed CI before, this one is on me. Coverage decrease is on you though

Issue #492, opened 2022-07-28
t_before and t_after not aligned correctly in brainbox.io.one.load_trials_df 


KaiNylund:
Hi! The documentation for brainbox.io.one.load_trials_df says it returns trials from t_before seconds before stim_on to t_after seconds after feedback, but in the source code it looks like t_before and t_after are both aligned to the same event (defaulting to stim_on) in the two lines below:
```
...
trialsdf['trial_start'] = trialsdf[align_event] - t_before
trialsdf['trial_end'] = trialsdf[align_event] + t_after
...
```

**To Reproduce**
```
trdf = bbone.load_trials_df(eid, maxlen=2., t_before=0.6, t_after=0.6,
                                      wheel_binsize=0.02, ret_abswheel=False,
                                      ret_wheel=True)
print(trdf["stimOn_times"] - trdf["trial_start"])
print(trdf["trial_end"] - trdf["stimOn_times"])                                  
```
All trials in the returned dataframe will be 0.6 seconds from trial start to stim onset and 0.6 seconds from stim onset to trial end

**Expected behavior**
All trials should be 0.6 seconds from trial start to stim onset and 0.6 seconds from feedback to trial end

PR #493, opened 2022-07-28
Fixing alignment bug in load_trials_df

KaiNylund:
Fix for: https://github.com/int-brain-lab/ibllib/issues/492


dbirman:
@oliche @mayofaulkner is this not going to be merged? It's annoying keeping my own separate version of this just for this purpose, if it isn't going to be merged can we make a separate function that does this properly?

mayofaulkner:
Hi @dbirman, this trials_df function is actually going to be removed from brainbox soon #529 as it has been causing confusion and has some errors. The new recommended way to load the trials data is using the [SessionLoader](https://int-brain-lab.github.io/iblenv/_autosummary/brainbox.io.one.html#brainbox.io.one.SessionLoader). If you need help changing your code to use this let me know and we can work on it together

PR #499, opened 2022-08-08
Adding event average resampling utils to brainbox.task.resampling_utils

KaiNylund:
Functions and examples for the full trial event averaging pipeline are stored in int-brain-lab/ibl-avg-trial-viz


k1o0:
Hello, thanks for doing this.  The tests are failing because you need to flake your functions.  You can see the errors [here](https://github.com/int-brain-lab/ibllib/runs/7718266039?check_suite_focus=true).  You can run flake locally by running `flake8 ./` from within your env, inside the ibllib repo.

PR #506, opened 2022-09-02
Made FDR correction optional for statistical testing

guidomeijer:
In the reproducible ephys taskforce we found that the Benjamin-Hochmann FDR correction from multiple testing can be quite restrictive. So I made the FDR correction optional in the Brainbox statistical tests.


oliche:
Thanks Guido ! 
I've cherry picked this commit to go in a bigger merge, it should be on develop soon

PR #509, opened 2022-09-12
MRI toronto atlas

dbirman:
- Adding a new atlas transformation (which we are comparing against Needles)
- Adding a parameter to xyz2ccf to allow setting the mode parameter of xyz2i


oliche:
Thanks for the contribution ! I'll add it in with the gene expression release in the next few days !

PR #525, opened 2022-10-17
Raise error when saving an empty dataset

k1o0:
Mayo and I noticed that for sessions where there are no camera TTLs, an empty camera.times dataset is saved.  As we weren't monitoring the camera QC closely we didn't notice these sessions.  I've modified the BaseExtractor class to raise a ValueError if the data being saved is empty.  The test session < v5.0.0 is actually empty. I don't know if this was always the case but now I assert that the extractor raises a ValueError for for this session.


k1o0:
No one raised any concerns to I'll merge

PR #527, opened 2022-10-24
typo fix

nsteinme:


k1o0:
Thanks for taking the time! It may take some time to propagate as we manually regenerate the docs every once in a while.

PR #530, opened 2022-10-24
Switch to previous value interpolation of position

berkgercek:
The current implementation of `brainbox.behavior.wheel.interpolate_position` relies on linear interpolation of a given point when resampling the encoder to a fixed frequency. This causes a given point at time `t` to be interpolated using information from both timestep `t - dt_1` and `t + dt_2` for some arbitrary deltas relative to the previous and next sampled points. This is an issue for causality.

More concretely, it results in some pretty crazy colinearities in the GLM, and I really need that fixed so I can finish a run for the BWM.

### DIFF
--- a/brainbox/behavior/wheel.py
+++ b/brainbox/behavior/wheel.py
@@ -24,7 +24,7 @@ ENC_RES = 1024 * 4  # Rotary encoder resolution, assumes X4 encoding
 WHEEL_DIAMETER = 3.1 * 2  # Wheel diameter in cm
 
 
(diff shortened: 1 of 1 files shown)


berkgercek:
Failing tests, likely because this breaks behavior w.r.t. test fixtures with a known interpolated signal. I'm currently running my glm code on the cluster off of this branch, but let me know when the change is merged!

oliche:
Yes this is all the QC tests that are changed because of this. @k1o0 recommendation was to inspect visually and change the test when validated.
Happy to do it when I'm back, I don't think this is a blocking issue for now.

PR #532, opened 2022-10-28
Paw QC metrics

themattinthehatt:
New entry in DLC QC pipeline that checks for proportion of nans in paw traces around trial windows.


juhuntenburg (commented):
Looks good to me! We might just have to add a test, the integration tests and coverage didn't run on this PR. Looking into it :)

juhuntenburg:
I don't understand why this still fails, they tasks pass locally for me ...

PR #542, opened 2022-11-24
update check copy complete for cases where experiment description file exists

mayofaulkner:


k1o0:
I think you'll need to check for a non-empty `_devices` folder too.  The copy script will copy the data then update the main description file and delete the individual file in _devices.  When the _devices folder is empty it means the session is ready to extract.  Supposedly this will be more robust (I'm not yet convinced though).

mayofaulkner:
Which copy script is this? I see this as an interim solution until all the details of the devices and creation of the experiment description file from different computers etc is finalised and implemented. I think this is still a while off (maybe not). In the meantime I want to be able to copy over sessions that have an experiment description file and not only rely on the task protocol to decide whether the copy has been complete

PR #552, opened 2023-01-03
Atlas

oliche:


k1o0:
FYI the CI was initially failing because the iblscripts atlas branch was so old that `01_run_coveralls.sh` didn't exist on that branch so after checking out the branch the CI couldn't run the test script!

PR #613, opened 2023-05-30
Atlas fix

oliche:
A few minor fix and a larger one concerning the volume computation.

By default we reverted to a simple sum of voxels per acronym, leaving the task of computing the cumulative volume up the region hierarchy to the user.

### DIFF
--- a/ibllib/io/session_params.py
+++ b/ibllib/io/session_params.py
@@ -470,7 +470,7 @@ def prepare_experiment(session_path, acquisition_description=None, local=None, r
         return
     # Determine if user passed in arg for local/remote subject folder locations or pull in from
     # local param file or prompt user if missing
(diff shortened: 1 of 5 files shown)


ibllib/io/session_params.py
```diff
@@ -470,7 +470,7 @@ def prepare_experiment(session_path, acquisition_description=None, local=None, r
         return
     # Determine if user passed in arg for local/remote subject folder locations or pull in from
     # local param file or prompt user if missing
-    params = misc.create_basic_transfer_params(transfers_path=local, remote_data_path=remote)
```
oliche, line 473: @k1o0 seems like a typo to me could you please have a look ?

k1o0: Good catch! Yes it does appear to be a typo.

PR #614, opened 2023-06-08
Aggregate docs fix

k1o0:
Fix for nan filling wrong columns when merging training and trials aggregates.  I sorted by both session start time and trial start time, then 

I also modified the trial loading changes that Gaelle made because it's not clear to me why we should recommend the SessionLoader for this.  I understand that it's easier for dealing with resolved vs unresolved probes but otherwise I don't know what the benefit would be (if there is one we should explain it in the docs!)

Because we use load_object a lot in the documentation and code I don't want to explicitly recommend against it without explanation, otherwise it's confusing for users.  If there is good reason to we can add back the explicit recommendation with an explanation.

### DIFF
--- a/.github/workflows/ibllib_ci.yml
+++ b/.github/workflows/ibllib_ci.yml
@@ -18,7 +18,12 @@ jobs:
       max-parallel: 4
       matrix:
         os: ["windows-latest", "ubuntu-latest"]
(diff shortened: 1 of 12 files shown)


k1o0:
No response for over a week so I'm going to merge :)

PR #620, opened 2023-06-19
Deprecate pyschofit in favour of psychofit pip package

k1o0:
psychofit is now a pip package with the correct spelling.


k1o0:
NB: Still need to remove from https://int-brain-lab.github.io/iblenv/notebooks_external/loading_trials_data.html#Useful-modules 
@mayofaulkner Maybe there are other places in the documentation that need changes?

PR #621, opened 2023-06-20
make sure rest cache is cleared before searching for already register…

mayofaulkner:
…ed datasets


k1o0:
I'm not sure that works :/ I think you need `no_cache=True` rather than `query_type`.

PR #622, opened 2023-06-21
Deprecate `brainbox.processing.bincount2D`

chris-langfield:
From the discussion in https://github.com/int-brain-lab/iblutil/pull/19 we are going to move `bincount2D` to `iblutil`. 

This replaces `brainbox.processing.bincount2D` with a call to the imported version from `iblutil` and adds a warning. 

Also updates the `requirements` to `iblutil` 1.6.1, which contains the replaced method.

### DIFF
--- a/brainbox/processing.py
+++ b/brainbox/processing.py
@@ -8,6 +8,10 @@ import pandas as pd
 from scipy import interpolate, sparse
 from brainbox import core
 from iblutil.util import Bunch
+from iblutil.numerical import bincount2D as _bincount2D
+import logging
+
+_logger = logging.getLogger('ibllib')
 
 
 def sync(dt, times=None, values=None, timeseries=None, offsets=None, interp='zero',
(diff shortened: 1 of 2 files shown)


k1o0 (changes requested):
Looks good. Just change the requirements to a minor release.  For the other rigs I've been raising both a DeprecationWarning and a logger warning in case users turn off the ibllib logger but that's overkill to be honest.  We also usually include a print statement of the traceback because it's sometime very difficult to find the call to that function, like this: https://github.com/int-brain-lab/ibllib/blob/381d6485279e2364ef92bcede07e6c8ecaff0d52/brainbox/behavior/wheel.py#L88-L89

It's also super annoying though (which may be a good thing). Personally I don't think it's particularly necessary. 

Finally, Olivier adds something that breaks the test as a reminder to remove the deprecated function e.g. https://github.com/int-brain-lab/ibllib/commit/2f6a13e062a28a3f9e0dade670cecead38c7c231. I just set a reminder in Slack once the branch has been merged for release.

PR #623, opened 2023-06-26
Deprecate velocity, velocity_smothed, and last_movement_onset

k1o0:
* deprecated `velocity`, `velocity_smothed`, and `last_movement_onset`
* `get_movement_onset` function
* updated wheel documentation

**NB**: Replacing velocity_smoothed with velocity_filtered in the wheel alignment code significantly changes the cross correlation output.  We'll need to check if it's an improvement.


CodyCBakerPhD:
Just curious - what is the reason for deprecating those functions?

We ran them and included the one-off output in the NWB version of the initial BWM release - are you recommending we remove it from there as well?

k1o0:
>Just curious - what is the reason for deprecating those functions?
>
>We ran them and included the one-off output in the NWB version of the initial BWM release - are you recommending we remove it from there as well?

The velocity function was never really used by us (sorry for the miscommunication on that).  It simply returns the diff of the wheel position.  To reduce noise caused by imprecision in the recording, we filter the signal.  We'd advise against using the unfiltered velocity for analysis.

unnamed
*The green line represents the output of velocity_filtered.

If possible, it would be worth replacing the velocity with the filtered velocity version in the data.

PR #652, opened 2023-09-26
Density gain option

chris-langfield:
Adds `gain` option to the `Density` plot


mayofaulkner (commented):
Looks good to me. So if both gain and vmax and vmin are given then gain will take precedence? Is this the behavior you want?

chris-langfield:
Yeah exactly. I figured the method level argument should override the ones passed to `imshow`

PR #655, opened 2023-10-05
add @sleepless decorator

bimac:


ibllib/pipes/misc.py
```diff
@@ -1147,13 +1147,52 @@ class WindowsInhibitor:
...
+    @staticmethod
+    def uninhibit(quiet: bool = False):
+        if quiet:
+            log.debug("Allowing Windows from going to sleep")
+        else:
+            print("Allowing Windows to go to sleep")
+        WindowsInhibitor._set_thread_execution_state(WindowsInhibitor.ES_CONTINUOUS)
+
+
+def sleepless(func: Callable[..., Any]) -> Callable[..., Any]:
```
oliche, line 1173: Is there anyway to test the function (at least the syntax and arguments ? )
Even on linux you can probably decorate a dummy function and run it just to see that it returns the proper number of arguments and that the syntax is correct. You will at least get coverage from this.

oliche (changes requested):
Maybe just add a simple test (even in Linux) that decorates a dummy sum function.
This is just to get coverage so if some arguments are refactored we get notified !

bimac:
Added a basic test ...

k1o0 (approved):

PR #660, opened 2023-10-11
Probe geometry doc

GaelleChapuis:
Added to doc page on loading raw ephys data

### DIFF
--- a/examples/loading_data/loading_raw_audio_data.ipynb
+++ b/examples/loading_data/loading_raw_audio_data.ipynb
@@ -1,0 +1,168 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "5683982d",
+   "metadata": {},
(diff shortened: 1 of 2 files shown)


oliche (changes requested):
Olà, thanks for the contribution a few things to change:
- channel geometry and histology are two different things and should be separate examples. The former is raw data and just `lateral_um` and `axial_um` relative to the probe tip while the latter is full histology. This is a distinction that seems subtle but has large implications
- regarding the full file: `ibllib.io.spikeglx` and `ibllib.dsp.voltage` are deprecated and are in the `ibl-neuropixel` repository, in `spikeglx` and `neurodsp.voltage` respectively.

GaelleChapuis:
`lateral_um` and `axial_um` are now used in the documentation.
Imports are done via `spikeglx` and `neurodsp`

Issue #715, opened 2024-01-18
Remove usage of deprecated extractor type in training_status


k1o0:
Currently the training status module is using the deprecated extractor type function(s) to determine the task type. Instead it should use protocol2extractor or get_bpod_extractor_class which avoids the extraneous 'extractor type' string.

PR #721, opened 2024-01-25
Training use extractor map

mayofaulkner:


k1o0:
Resolves https://github.com/int-brain-lab/ibllib/issues/715

PR #868, opened 2024-10-17
Resolves #865

### LINKED ISSUE #865: Support mixed enums in MesoscopePreprocess._consolidate_exptQC
Currently `MesoscopePreprocess._consolidate_exptQC` will raise if the exptQC.mat files have a different enumeration. This is unnecessary as the enumerations are entirely independent for each bout and a new hybrid enumeration can be generated instead of raising.  This would make things much easier for experimenters as they don't have to remember their QC labels between between bouts.

k1o0:
I've refactored the function and added the following test:
```
exptQC = [
    {'frameQC_names': np.array(['ok', 'PMT off', 'galvos fault', 'high signal'], dtype=object),
     'frameQC_frames': np.array([0, 0, 0, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 4])},
    {'frameQC_names': np.array(['ok', 'PMT off', 'foo', 'galvos fault', np.array([])], dtype=object),
     'frameQC_frames': np.array([0, 0, 1, 1, 2, 2, 2, 2, 3, 4])},
    {'frameQC_names': 'ok',  # check with single str instead of array
     'frameQC_frames': np.array([0, 0])}
]
```
This will produce a new enumeration that looks like this:
```
   qc_values     qc_labels
0          0            ok
1          1       PMT off
2          2  galvos fault
3          3   high signal
4          4           foo
5          5       unknown
```
So the behaviour is this:

- ok is always 0, the other values are purely determined by the order of frameQC names first encountered;
- if values correspond to different names across bouts this will not result in an error
- all values that don't correspond to a label will be assigned the 'unknown' label
- an empty str or empty array will be assigned the same 'unknown' label instead of raising
- if names is simply an 'ok' str (instead of an array) this is handled as usual

### DIFF
--- a/ibllib/pipes/mesoscope_tasks.py
+++ b/ibllib/pipes/mesoscope_tasks.py
@@ -486,23 +486,42 @@ class MesoscopePreprocess(base_tasks.MesoscopeTask):
         numpy.array
             An array of frame indices where QC code != 0.
         """
-
-        # Merge and make sure same indexes have same names across all files
-        frameQC_names_list = [e['frameQC_names'] for e in exptQC]
-        frameQC_names_list = [{k: i for i, k in enumerate(ensure_list(f))} for f in frameQC_names_list]
-        frameQC_names = {k: v for d in frameQC_names_list for k, v in d.items()}
-        for d in frameQC_names_list:
-            for k, v in d.items():
-                if frameQC_names[k] != v:
-                    raise IOError(f'exptQC.mat files have different values for name "{k}"')
-
-        frameQC_names = pd.DataFrame(sorted([(v, k) for k, v in frameQC_names.items()]),
-                                     columns=['qc_values', 'qc_labels'])
-
+        # Create a new enumeration combining all unique QC labels.
+        # 'ok' will always have an enum of 0, the rest are determined by order alone
+        qc_labels = ['ok']
+        frame_qc = []
+        for e in exptQC:
+            assert e.keys() >= set(['frameQC_names', 'frameQC_frames'])
+            # Initialize an NaN array the same size of frameQC_frames to fill with new enum values
+            frames = np.full(e['frameQC_frames'].shape, fill_value=np.nan)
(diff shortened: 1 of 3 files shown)


ibllib/pipes/mesoscope_tasks.py
```diff
@@ -486,23 +486,42 @@ def _consolidate_exptQC(exptQC):
...
+        # If any NaNs left over, assign 'unknown' label
+        if (missing_name := np.isnan(frame_qc)).any():
+            try:
+                i = qc_labels.index('unknown')
+            except ValueError:
+                i = len(qc_labels)
+                qc_labels.append('unknown')
+            frame_qc[missing_name] = i
+        frame_qc = frame_qc.astype(np.uint32)  # case to uint
+        bad_frames, = np.where(frame_qc != 0)
```
samupicard, line 521: the frames that should be excluded from suite2p cell detection are not necessarily all `frame_qc != 0`. The acquirer will have written a separate badframes.mat vector specifying which frames suite2p should be excluding from analysis. So probably best to not impute & return the bad_frames variable here at all, and instead load it from disk?

samupicard: if loaded from disk, this badframes array will also need to be consolidated (the indexing of the badframes.mat vectors is relative to each bout)

samupicard (commented):
the intended behavior of _consolidate_exptQC looks good to me! Thanks for adding a test as well

ibllib/pipes/mesoscope_tasks.py
k1o0: I can confirm that the bad_frames var returned by this method is not at all used. The badframes.mat files are loaded afterward, and there is an assertion that all badframes indices are a subset of the non-zero frameQC indices:
https://github.com/int-brain-lab/ibllib/blob/c4418cb527cf9eea10c198638819ad640c000144/ibllib/pipes/mesoscope_tasks.py#L773-L787

PR #890, opened 2024-11-29
Add an option to run tasks so they raise an error during the run

oliche:
Currently tasks catch errors and output the traceback.
For some manual processing and or scripts, it is useful to halt execution and raise during the task.

The `on_error` parameter at the init stage is set to `continue` by default, and can be set to `raise` to achieve this behaviour.


oliche:
Look at run_alyx tasks that may benefit from using this

k1o0:
> Look at run_alyx tasks that may benefit from using this

https://github.com/int-brain-lab/ibllib/blob/e110a244db60f67cd47d1e01c10a8faec132885c/ibllib/pipes/tasks.py#L912-L914
Interestingly this cleans up the task before re-raising

PR #903, opened 2025-01-09
Update README.md

kdharris101:
Changed installation documentation link since docs.internationalbrainlab.org was down


oliche:
Thanks !
We did implement a check and alerting on the availability of the domain, so this shouldn't be an issue going forward.

PR #905, opened 2025-01-13
propsed fix for spikeglx streamer failing on SDSC because of wrong uuid in file name

grg2rsr:
previously, the Streamer was instantiated with a absolute path of the meta file (also not ideal in the context of the Streamers constructors call signature, where the first argument is named `sglx_file` and a dedicated argument `meta_file` exists. [link](https://github.com/int-brain-lab/ibl-neuropixel/blob/185489b481df43c78c23e44f52c1befa3fb64bc3/src/spikeglx.py#L56)). While this looks like a bug, it is recovered there in the following lines under normal (=non-SDSC) usage as this line 
```
meta_file = meta_file or _get_companion_file(sglx_file, ".meta")
```
recoveres the correct file paths.

However, this breaks on the SDSC as the file paths an uuid, and when the corresponding file path for the meta for the bin (or vice versa) are constructed, the uuid is wrong.

The proposed patch checks if the `.meta` filename contains an uuid, and if so, finds the correct uuid for the `.cbin` file, constructs absolute paths for both. Either way, uuid or not, the `Streamer` is now instantiated with  both paths for `.meta` and `.bin` (or `.cbin`).

### DIFF
--- a/brainbox/io/spikeglx.py
+++ b/brainbox/io/spikeglx.py
@@ -80,32 +80,33 @@ def extract_waveforms(ephys_file, ts, ch, t=2.0, sr=30000, n_ch_probe=385, car=T
         ch = np.asarray(ch)
         ch = ch.reshape((ch.size, 1)) if ch.size == 1 else ch
         if np.any(ch < 0) or np.any(ch > n_ch_probe):
-            raise Exception(
-                'At least one specified channel number is impossible. '
-                f'The minimum channel number was {np.min(ch)}, '
-                f'and the maximum channel number was {np.max(ch)}. '
-                'Check specified channel numbers and try again.'
-            )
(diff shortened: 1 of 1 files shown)


grg2rsr:
thinking more about this, this could also be patched in a different way, where the uuid completing functionality is added to the `_get_companion_file()` function in [ibl-neuropixel/spikeglx](https://github.com/int-brain-lab/ibl-neuropixel/blob/185489b481df43c78c23e44f52c1befa3fb64bc3/src/spikeglx.py#L22), with a  `remove_uuid=True` as a default kwarg, which is then turned to `False` either here and passed through the `Streamer` constructor. This could cover future occurrences of this bug.

oliche (commented):
SDSC is probably the only place on Earth where we do not need the Streaming functionality ! :smile:

PR #927, opened 2025-02-07
CI test fixes

k1o0:
Changes in ONE v3.0b4 should allow UUID objects in both REST filters and PUT/PATCH data. These changes are for the Django query strings.


k1o0:
The last failing test here requires that we update the test database. This has been done however the new test dump needs to be copied to the cortexlab CI.  This is usually done by a script however `/var/www/alyx-test/data/reset_alyx.sh` is failing at the moment

PR #937, opened 2025-02-19
add options for SDSC and EC2 patchers

oliche:
During the patching of the wheel data, I needed some more control of patching using EC2 and SDSC.

- SDSC patcher operates on symlinks, and if one of the input files is modified, this directly impacts the original data. I have set an assertion that enforces distinct input and output datasets
- the EC2 patcher didn't check file hashes on checking local data, I've set it to true by default and provided an option to opt-out for bulky input data such as spike sorting. `task.run(hash_files=False)`

### DIFF
--- a/ibllib/oneibl/data_handlers.py
+++ b/ibllib/oneibl/data_handlers.py
@@ -16,7 +16,7 @@ from copy import copy
 from one.api import ONE
 from one.webclient import AlyxClient
 from one.util import filter_datasets
-from one.alf.path import add_uuid_string, session_path_parts, get_alf_path
+from one.alf.path import add_uuid_string, get_alf_path, ensure_alf_path
 from one.alf.cache import _make_datasets_df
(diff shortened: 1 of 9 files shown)


ibllib/oneibl/data_handlers.py
```diff
@@ -973,8 +974,9 @@ def setUp(self, task):
                     Path(self.root_path.joinpath(file_uuid)))
             except FileExistsError:
                 pass
-
         task.session_path = SDSC_TMP.joinpath(session_path)
+        assert self.getOutputFiles(session_path=task.session_path).shape[0] == 0, (
+            "On SDSC patcher, input files output files should be distinct from input files to avoid overwriting")
```
k1o0, line 979: Minor typo here. Also could you please add a comment about copying input files that are also output file in the future? Otherwise if we encounter this assertion in the future we may not remember the implications of breaking it

PR #1033, opened 2025-09-23
querying revisions with id will fail with new one release

oliche:


k1o0 (changes requested):
This is related to changing the endpoint so that ID is no longer the revision name, correct? If so this may affect other locations in ibllib. In this file I think line 642 would need updating.

PR #1066, opened 2025-12-08
PostLP task fix

grg2rsr:
this fixes Matt's problem:

(from slack):
_I have a troubleshooting question that may (or may not) have a simple answer. For the PostLP task I'm working with, the signature property looks like this:_
```
    @property
    def signature(self):
        return {
            'input_files': [(f'_ibl_{cam}Camera.lightningPose.pqt', 'alf', True) for cam in self.cameras] +
                           [(f'_ibl_{cam}Camera.times.npy', 'alf', True) for cam in self.cameras] +
                           ...
                           ...
                           [('_ibl_trials.table.pqt', self.trials_collection, True),
                            ('_ibl_wheel.position.npy', self.trials_collection, True),
                            ('_ibl_wheel.timestamps.npy', self.trials_collection, True)],
```
_the problem I'm facing is that the trials_collection for some (steinmetz) sessions is alf, while for others it's alf/task_00 (and there may be others), and the current data handler (ec2) doesn't handle this gracefully_

and addresses Oliviers comment:
_Yet there will be another issue: if you are running from EC2, you need to download the experiment description file for this method to work properly. For now you can add this to the input signature of your tasks, but we were actually discussing today that we should add this for all of the tasks, ideally using the same pattern of adding this in the base classes._

by adding the `_ibl_experiment.description.pqt` to the list of input files by extending the `get_signatures()` method. Not very pretty.

### DIFF
--- a/ibllib/pipes/base_tasks.py
+++ b/ibllib/pipes/base_tasks.py
@@ -279,7 +279,8 @@ class BehaviourTask(DynamicTask):
 
     def get_signatures(self, **kwargs):
         """ explicitly adding the experiment description file to the input files for all behavior tasks
-        This is pretty hacky but avoids having to modify all the child classes"""
+        There is a trade-off between this inheritance design and explicit file signatures for child classes"""
+
         super().get_signatures(**kwargs)
         self.input_files.append(('_ibl_experiment.description.pqt', self.collection, True))
         return super().get_signatures(**kwargs)
--- a/ibllib/pipes/base_tasks.py
+++ b/ibllib/pipes/base_tasks.py
@@ -277,8 +277,15 @@ class BehaviourTask(DynamicTask):
             raise ValueError('No trials data and/or extractor found')
         return trials_data
 
+    def get_signatures(self, **kwargs):
+        """ explicitly adding the experiment description file to the input files for all behavior tasks
+        This is pretty hacky but avoids having to modify all the child classes"""
+        super().get_signatures(**kwargs)
+        self.input_files.append(('_ibl_experiment.description.pqt', self.collection, True))
+        return super().get_signatures(**kwargs)
 
-class VideoTask(DynamicTask):
+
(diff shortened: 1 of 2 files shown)


ibllib/pipes/video_tasks.py
```diff
@@ -818,13 +818,14 @@ class PostLP(base_tasks.VideoTask):
...
         super().__init__(*args, **kwargs)
-        self.trials_collection = kwargs.get('trials_collection', 'alf')
+        self.trials_collection = self.get_task_collection()
 
     @property
     def signature(self):
         return {
             'input_files': [(f'_ibl_{cam}Camera.lightningPose.pqt', 'alf', True) for cam in self.cameras] +
                            [(f'_ibl_{cam}Camera.times.npy', 'alf', True) for cam in self.cameras] +
+                           ['_ibl_experiment.description.pqt', self.trial_collection, True] +
```
oliche: I think the common construct is not a list but a tuple, for example: `('*trials.rewardVolume.npy', self.output_collection, True),`, but I am not sure this has any impact.

ibllib/pipes/base_tasks.py
```diff
@@ -277,8 +277,15 @@ def _assert_trials_data(self, trials_data=None):
             raise ValueError('No trials data and/or extractor found')
         return trials_data
 
+    def get_signatures(self, **kwargs):
+        """ explicitly adding the experiment description file to the input files for all behavior tasks
+        This is pretty hacky but avoids having to modify all the child classes"""
```
oliche: I would rephrase the comment
```
        """ explicitly adding the experiment description file to the input files for all behavior tasks
        There is a trade-off between this inheritance design and explicit file signatures for child classes"""
 ```
 
 There are 8 child instances in ibllib alone, here both sides of the debate have a point.

oliche (approved):

PR #1071, opened 2026-01-08
small bugfix for loading experiment description files where protocols…

grg2rsr:
… are not a list

this snippet reproduces the problem:
```
from one.api import ONE
import ibllib.io.session_params as sess_params

one = ONE()
eid = "1c61be53-cb71-469c-b2a3-304f7bb5ec2f"
ed = one.load_dataset(eid, '*experiment.description')
sess_params.get_task_collection(ed)
```

### DIFF
--- a/ibllib/io/session_params.py
+++ b/ibllib/io/session_params.py
@@ -31,7 +31,7 @@ from itertools import chain
 from copy import deepcopy
 
 from one.converters import ConversionMixin
-from iblutil.util import flatten
+from iblutil.util import flatten, ensure_list
 from iblutil.io.params import FileLock
 from packaging import version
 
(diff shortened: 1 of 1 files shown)


grg2rsr:
fixed by patching the affected experiment description files

k1o0:
The loader has a patch function that fixes this issue but it only affects description files <= v0.1
https://github.com/int-brain-lab/ibllib/blob/265f7dda7a40be554eedcccde7f686e1299019dd/ibllib/io/session_params.py#L76-L79

What version were these experiment files before being loaded?

grg2rsr:
just had a look, the affected experiment description files do not have a `version` entry

PR #1075, opened 2026-01-30
make pose qc plotting robust to non-standard framerates

themattinthehatt:


brainbox/behavior/dlc.py
```diff
@@ -202,21 +204,22 @@ def get_pupil_diameter(dlc):
...
     :param std_thresh: threshold (in standard deviations) beyond which a point is labeled as an outlier
     :param nan_thresh: threshold (in seconds) above which we will not interpolate nans, but keep them
                        (for long stretches interpolation may not be appropriate)
+    :param fr: framerate (frames per second); if None, default value based on camera string used
     :return:
     """
     # set framerate of camera
     if camera == 'left':
-        fr = SAMPLING['left']  # set by hardware
+        fr = fr or SAMPLING['left']  # set by hardware
```
oliche: Here maybe more explicit to `fr if fr is None else SAMPLING['left']` so that there is no ambiguity if the frame rate turns out to be 0 for some reason.

brainbox/behavior/dlc.py
```diff
@@ -202,21 +204,22 @@ def get_pupil_diameter(dlc):
...
         window = 31  # works well empirically
     elif camera == 'right':
-        fr = SAMPLING['right']  # set by hardware
+        fr = fr or SAMPLING['right']  # set by hardware
```
oliche: same same

oliche (approved):
From what I understand we want to be free of the hard-coded framerates as constants. LGTM. Thanks !

brainbox/behavior/dlc.py
themattinthehatt: @oliche do you mean
```
fr = fr if fr is not None else SAMPLING['left']
```
(i.e. *not* None)

PR #1084, opened 2026-02-11
paw segmentation with lightning action

themattinthehatt:


ibllib/pipes/video_tasks.py
```diff
@@ -960,3 +960,212 @@ def _run(self, overwrite=True, run_qc=True, plot_qc=True):
                 self.status = -1
 
         return output_files
+
+
+class LightningAction(base_tasks.VideoTask):
```
oliche, line 999: Maybe add a docstring giving some context about the task and its usage. Also if there is anything specific about the litaction environment.

ibllib/pipes/video_tasks.py
```diff
@@ -960,3 +960,212 @@ def _run(self, overwrite=True, run_qc=True, plot_qc=True):
...
+                           [('_ibl_wheel.position.npy', self.trials_collection, False),
+                            ('_ibl_wheel.timestamps.npy', self.trials_collection, False)],
+            'output_files': [
+                (f'_ibl_{cam}Camera.pawstates.pqt', 'alf/lightningaction', True) for cam in self.cameras if cam != 'body'
+            ]
+        }
+
+        return signature
+
+    def _check_env(self):
```
oliche, line 1042: This is good for now, but I think we will move away from this with the reliability WG. Just FYI

ibllib/pipes/video_tasks.py
```diff
@@ -960,3 +960,212 @@ def _run(self, overwrite=True, run_qc=True, plot_qc=True):
...
+
+        # If all results exist and overwrite is False, skip computation
+        expected_outputs_present, expected_outputs = self.assert_expected(self.output_files, silent=True)
+        if overwrite is False and expected_outputs_present is True:
+            actual_outputs = expected_outputs
+            return actual_outputs
+
+        # Else, loop over videos
+        actual_outputs = []
+        for label in self.cameras:
```
oliche, line 1080: For ephys we found that it was easier to create a single task for each probe. I am not sure this is something you had ever considered ? It is quite a refactoring though...

This would simplify the code below by avoiding having to catch the exception for each camera and facilitates backfill.

ibllib/pipes/video_tasks.py
```diff
@@ -960,3 +960,212 @@ def _run(self, overwrite=True, run_qc=True, plot_qc=True):
...
+
+    @property
+    def signature(self):
+        signature = {
+            'input_files': [(f'_ibl_{cam}Camera.lightningPose.pqt', 'alf', True) for cam in self.cameras if cam != 'body'] +
+                           [(f'_ibl_{cam}Camera.times.npy', 'alf', True) for cam in self.cameras if cam != 'body'] +
+                           [('_ibl_wheel.position.npy', self.trials_collection, False),
+                            ('_ibl_wheel.timestamps.npy', self.trials_collection, False)],
+            'output_files': [
+                (f'_ibl_{cam}Camera.pawstates.pqt', 'alf/lightningaction', True) for cam in self.cameras if cam != 'body'
```
oliche, line 1036: I'll submit this to the data polizei today at the DAWG meeting, I'll update here.

ibllib/plots/figures.py
```diff
@@ -922,3 +933,282 @@ def pose_qc_plot(session_path, one=None, device_collection='raw_video_data',
...
+def pawstates_qc_plot(
+        session_path, one=None,
+        camera='left', paw='paw_l', tracker='lightningPose',
+        device_collection='raw_video_data', trials_collection='alf',
+):
+    """
+    Creates pawstates behavioral QC plot.
+    Data is searched first locally, then on Alyx. Panels that lack required data are skipped.
+
+    Required data to create all panels:
```
oliche: Super useful, thanks !

ibllib/plots/figures.py
```diff
@@ -922,3 +933,282 @@ def pose_qc_plot(session_path, one=None, device_collection='raw_video_data',
...
+    :param one: ONE instance, if None is given, default ONE is instantiated
+    :param camera: Camera view ('left' or 'right')
+    :param paw: Paw identifier ('paw_l' or 'paw_r')
+    :param tracker: Tracker type (e.g., 'dlc', 'lightningPose')
+    :param device_collection: Collection name for video data
+    :param trials_collection: Collection name for trials data
+    :returns: Matplotlib figure
+    """
+
+    one = one or ONE()
```
oliche: Here in terms of design it would be nice to have a function that specifically creates the plots from a simple folder architecture, without using one. That decouples the loading from the plotting code.

This way you can hit 2 birds with one stone, not only you have the QC plots that could possibly generalize beyond our infrastructure, but you can also wrap the function with ONE to run in the task infrastructure.

oliche (changes requested):
Hi Matt, thanks !

So we're doing paw segmentation now 🤩 .

I gave you some design advice, that you may implement or not depending your load:

- splitting the tasks per camera: seems a large refactoring, that would concern the whole video pipeline so maybe not right now
- having a stand-alone QC function that decouples loading from plotting

The former is much more manageable, but none of them is an absolute requirement for merging.

I would have one requirement though, could you please write a unit-test that goes through the task ? I'm not sure what is the minimum viable dataset you'd need for a test, would a few frames suffice ? This would make the potential refactoring by someone else much easier.

ibllib/pipes/video_tasks.py
themattinthehatt: I already spoke with @k1o0! This dataset type should be good to go, and I've already run the LightningAction task on a couple hundred sessions :)

ibllib/plots/figures.py
themattinthehatt: sounds good, I'll work on this

themattinthehatt:
Thanks for the feedback! 

Re: splitting the tasks per camera: I'll leave this for now; it's easy enough to run a task on the left cam, then rerun on the right, so the two are functionally decoupled already anyways. Regarding the entire video pipeline, the only task that really requires more than a single camera is PostLP, which uses both left/right cams to compute lick times.

Re: figure refactoring, yes this is a good idea, will do.

Re: unit tests, will also do this. Do you have a good example of a task unit test I can look at?

ibllib/pipes/video_tasks.py
oliche: Yes I had questions about the subfolder, not the name. We just discussed it at the data meeting and this is fine as is.

oliche (approved):

PR #1096, opened 2026-04-07
moving photometry related code out of ibllib

grg2rsr:
main changes:
this PR: moves all photometry related code out of ibllib
 - `pipes.neurophotometrics` (all task related code) now is in ibl-photometry `0.3.0`
 - ibl-photometry>=0.3.0 is now a dependency
 - `dynamic_pipeline.py` imports tasks from ibl-photometry
 - `PhotometrySessionLoader` (used to be in brainbox.io.one, in the `photometry-integration` branch) is now in `iblphotometry.fpio` <- this change has the largest influence on users, which need to change the imports in their code. However, since there is currently a quite limited set of users, this is manageable
 - removed photometry data loading example notebook

reasoning for these changes:
1) the current extraction process is happening with different ibllib branches (`3.5rc1` on whiterussian) and `photometry-integration` on parede which have diverged, bugfixes to the extraction process involve distributed edits across those branches, as well as keeping everything in line with the actual photometry repository. Additionally, the dockerization on parede requires rebuilding the containers for `photometry-integration` after each step. Moving all extraction related code to `ibl-photometry` will simplify development.
2) as discussed in the data architecture WG, this will shrink down `ibllib` and reduce overloading of functionality. All base classes in ibl-photometry import from `ibllib`, vice versa `ibllib` only depends on `iblphotometry` because of the dynamic pipeline creation being specified here.


examples/loading_data/loading_photometry_data.ipynb
oliche, line 1: If we want to move this notebook in ibl-photometry, then the iblenv branch of the documentation needs to pull in ibl-photometry and import the notebooks from there. This should be done before the file is removed !

We can also keep this file here. Either would work, but in the former case this can only be deleted after changes have been made to iblenv.

requirements.txt
```diff
@@ -33,3 +33,4 @@ psychofit
 slidingRP>=1.1.1  # steinmetz lab refractory period metrics
 pyqt5
 ibl-style
+ibl-photometry>=0.3.0
```
oliche, line 36: There is an added dependency here that is pywavelets, which depends only on numpy. I think this is acceptable ! 🤞

oliche (changes requested):

examples/loading_data/loading_photometry_data.ipynb
grg2rsr: added back in

requirements.txt
grg2rsr: pywavelets is only used for synthetic data generation, which we are also currently not really using. Very fine with dropping it

PR #1126, opened 2026-07-20
Feature/tidy trials table loader

grg2rsr:
added loading tidy trials tables functionality to the `SessionLoader` - default is previous ONE behavior, add method originally authored by Heberto from CatalystNeuro to convert the trials table to tidy format. Loading behavior of the SessionLoader can be modified by a kwarg, i.e. `session_loader.load_trials(style='tidy')` will load the trials table in tidy format

### DIFF
--- a/brainbox/io/one.py
+++ b/brainbox/io/one.py
@@ -5,7 +5,6 @@ import gc
 import logging
 import re
 import copy
-import math
 from pathlib import Path
 from collections import defaultdict
 from typing import Literal
@@ -1425,7 +1424,7 @@ class SessionLoader:
         self.one.wildcards = True
         self.data_info.loc[self.data_info['name'] == 'trials', 'is_loaded'] = True
         if style == 'tidy':
-            self.trials = self._apply_tidy_transformations(self.trials)
+            self.trials = self.apply_tidy_transformations(self.trials)
 
     def load_wheel(self, fs=1000, corner_frequency=20, order=8, collection=None):
         """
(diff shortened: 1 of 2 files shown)


grg2rsr:
on caveat currently:
`session_loader.load_session_data(trials=True)` registers the `.load_trials()` method with default kwargs (also no control over revision) - potential solution would be to pass in a dict with loader specific kwargs and have register callables using partial ... not sure if this is overkill or if anyone would ever want to use this. I could see this make sense though for revisions ...

oliche (changes requested):
We need to add a test, even a simple one from a dummy fixture just to check the signature and the calls (the full part is probably covered in the integration test). 

Then optionally, I would refactor this to have the conversion vectorized so we can run on a larger scale (ie. aggregate trials tables). Right now the list comprehensions and loop over trials will be unnecessarily long !

grg2rsr:
**Performance: vectorized `SessionLoader.apply_tidy_transformations`**

The tidy transformation had three per-row Python paths, now replaced with vectorized numpy/pandas equivalents (commit a0cfad9c):

- **`gabor_stimulus_side` / `gabor_stimulus_contrast`** — were two Python list-comprehensions calling a per-row helper via `zip`; now `np.select` on `notna()` masks plus `left.fillna(right) * 100`.
- **`block_index`** — was an explicit `for` loop with `math.isclose` per trial; now `np.cumsum(~np.isclose(...))` on the shifted array.
- **`block_type`** — was `.apply()` with a per-row Python function; now `np.select` on `isclose`.

**Timing:** ~1.75 ms/call for a 1000-trial session. Negligible on a single session, but the removed per-row interpreter overhead matters when the loader runs across hundreds/thousands of sessions in batch analyses.

oliche (approved):
Great, this looks much better once vectorized. 

I had this reviewed by an agent, and one puzzling thing resurfaced: 

>   When both contrastLeft and contrastRight are populated (handled defensively in lines         
>   1684-1686), the code correctly determines which side has the larger contrast and sets        
>   gabor_stimulus_side accordingly. However, the contrast value extraction on line 1688 uses    
>   left.fillna(right), which always prefers contrastLeft regardless of which side was           
>   selected.                                                                                    
>                                                                                                
>   Example of the bug:                                                                          
>                                                                                                
>   contrastLeft = 0.25, contrastRight = 0.5                                                     
>   # Result: side='right' ✓, contrast=25.0 ✗ (should be 50.0)                                   
>   Fix required: The contrast extraction must use the value from the selected side, not         
>   always from left.  

Clearly this doesn't apply to the IBL task, but should we enforce some kind of validation here or address this ? I'll let you decide on how to approach this !

grg2rsr:
thx for the review. I think since this code will only ever convert IBL task trial tables, the bug should never be triggered ...

PR #1130, opened 2026-08-03
replace DLC with LP in data loading docs

themattinthehatt:
Current `ibllib` data loading docs still reference `"dlc"` pose estimates; this PR updates the docs to reference the updated `"lightningPose"` estimates.

I have also verified that the reference eid in the example does in fact contain Lightning Pose estimates.


oliche (commented):
LGTM !

PR #1132, opened 2026-08-18
Migrate packaging to pyproject.toml

oliche:
## Summary
- Replace `setup.py` with `pyproject.toml` (PEP 621, setuptools backend), matching the style used by iblatlas/iblutil in this workspace
- Go all-in on uv: dependencies now inlined in `pyproject.toml` (same unpinned specifiers as before), `requirements.txt` removed, `uv.lock` added
- CI (`unit-tests.yml`) uses `astral-sh/setup-uv` + `uv sync`/`uv run` instead of pip
- Publish workflow uses `uv build` + `uv publish` instead of `build`/`twine` (reuses existing `PYPI_USERNAME`/`PYPI_PASSWORD` secrets)
- Version stays dynamic (from `ibllib/__init__.py`), entry point and package data preserved
- Build backend stays `setuptools`: `uv_build` only supports a single top-level package, but this project ships `ibllib`, `brainbox` and `examples` as separate top-level packages

## Test plan
- [x] Built sdist + wheel from this branch; metadata (version, dependencies, entry point, package data) matches prior `setup.py` output, and `brainbox`/`examples` are still present in the wheel
- [x] `pip install -e .` / `uv sync` continue to work


bimac:
why not go all in on `uv`?

https://docs.astral.sh/uv/guides/migration/pip-to-project/

- use `uv.lock` instead of `requirements.txt`
- use [setup-uv](https://docs.astral.sh/uv/guides/integration/github/) for the CI
- drop `setuptools` (replace with `uv_build`)
- [package and publish with uv](https://docs.astral.sh/uv/guides/package/)
- drop `twine`

compare:

- [bpod-core CI](https://github.com/int-brain-lab/bpod-core/blob/2ffcb5320db4f124c483345975d3dc1103f656fb/.github/workflows/publish.yaml#L33-L46)
- [bpod-core pyproject.toml](https://github.com/int-brain-lab/bpod-core/blob/2ffcb5320db4f124c483345975d3dc1103f656fb/pyproject.toml#L1-L3)

if you ever need a `requirements.txt` file for some deployment system, you can always create one on the fly:

``` bash
uv export --format requirements-txt > requirements.txt
```

oliche:
Went all-in on uv per your suggestion:

- Dependencies now live directly in `pyproject.toml` (same unpinned/loose specifiers as before, just inlined from `requirements.txt`, which is now removed).
- Added `uv.lock`.
- CI (`unit-tests.yml`) now uses `astral-sh/setup-uv` + `uv sync`/`uv run` instead of pip.
- Publish workflow now uses `uv build` + `uv publish` instead of `build`/`twine` (reusing the existing `PYPI_USERNAME`/`PYPI_PASSWORD` secrets via `UV_PUBLISH_USERNAME`/`UV_PUBLISH_PASSWORD`, no change needed on the PyPI side).

One thing I kept as-is: the build backend is still `setuptools`, not `uv_build`. I tested `uv_build` directly and it only supports a single top-level package per project (`module-name` rejects a list), but the published `ibllib` wheel ships three top-level packages (`ibllib`, `brainbox`, `examples`) - `brainbox` in particular is imported throughout the ecosystem. Switching to `uv_build` as-is would silently drop `brainbox`/`examples` from the wheel. Restructuring the repo to fit a single-package model would be a much bigger, separate breaking change, so I left that out of scope here.

bimac (commented):
It builds fine on my end.
A few things though (potentially for another time):

- the repository is not ruffed
- we should probably drop flake8
- for having several modules in a package see #1136

bimac (changes requested):
There actually are some issues with the packaging that came to light through automatic code review:

  1. **pyproject.toml:76 — namespace-package discovery silently turned on, shipping 79 extra files (medium)**                                                                                                                                                             
  In `pyproject.toml`, `[tool.setuptools.packages.find]` defaults to `namespaces = true`; the deleted `setup.py` used `find_packages(...)`, where it's off. Diffing the two builds: the wheel gains 79 files and the sdist 90 — `brainbox/docs/conf.py`, `brainbox/examples/*.py`, `examples/WIP/*`, `examples/archive/*`, `examples/ephys/*`, `examples/one/{behavior,ephys,histology}/*`, none of which have `__init__.py`. So `pip install ibllib` now drops example scripts and a Sphinx `conf.py` into site-packages and makes `brainbox.examples/brainbox.docs` importable. Any future top-level dir with a `.py` in it gets packaged too. Adding `namespaces = false` gives exact parity with the old build.                                                                                                                                                 
                                                                                                                                                                                                                                                                                
  2. **pyproject.toml:84 — Intel macOS declared required, but the lock has no Intel-macOS wheels for numba/llvmlite (medium)**                                                                                                                                                      
  `uv.lock` pins numba 0.67.0 / llvmlite 0.49.0 in a single unforked resolution, and both ship macOS arm64-only wheels. On an Intel Mac uv sync falls back to compiling llvmlite from sdist (needs a matching LLVM toolchain; realistically fails) while the manifest advertises that platform as supported. Also worth flagging: required-environments doesn't restrict the resolution universe — that's `[tool.uv]` environments — so 6215ee3f's stated goal isn't quite what the key does. Drop the `darwin/x86_64` entry, or switch to environments.           
                                                                                                                                                                                                                                                                                
  3. **pyproject.toml:9 — deprecated license table + unbounded setuptools will break the release build (low)**                                                                                                                                                                      
  Every build warns `'project.license' as a TOML table is deprecated — By 2027-Feb-18 ... remove deprecated calls`. Since `requires = ["setuptools>=64"]` has no upper bound and uv build pulls the newest setuptools into a fresh isolated env, `python-publish.yml` starts failing on a release tag the day the shim is removed. Use `license = "MIT"` + `license-files = ["LICENSE"]`.                                                                                                                                                                              
                                                                                                                                                                                                                                                                                
  4. **.github/workflows/unit-tests.yml:33 — uv sync without --locked (low)**                                                                                                                                                                                                       
  `uv sync --group dev` (and the implicit syncs in later uv run steps) will silently re-resolve and rewrite `uv.lock` on drift. A PR that edits dependencies without regenerating the lock passes CI green and a stale lock lands on master — the exact failure this migration is meant to prevent. Use `uv sync --locked --group dev`.                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                
  5. **.venv/ is not gitignored (low)**                                                                                                                                                                                                                                             
  The new setup path creates `.venv/` at the repo root. `.gitignore:14` has `venv/` and `ibllibenv/`, but git check-ignore `.venv` returns nothing — so a `git add -A` can commit the whole virtualenv.                                                                                     
                                                                                                                                                                                                                                                                                
  6. **pyproject.toml:20,42 — flake8 and pytest are still runtime dependencies (low)**                                                                                                                                                                                              
  Carried over verbatim from `requirements.txt`, so not a regression, but this PR is what introduces `[dependency-groups] dev` and is the natural place to move them. Currently every `pip install ibllib` pulls pytest and flake8>=3.7.8, which can collide with a downstream project's pinned pytest.

I think at least 1, 2, 5 and 6 should be fixed

oliche:
Merged latest \`develop\` in and addressed the packaging review:

1. Added \`namespaces = false\` to \`[tool.setuptools.packages.find]\` — verified with a clean \`uv build\`: wheel/sdist no longer ship \`brainbox/examples\`, \`brainbox/docs\`, etc. (326 files vs. the previous 405).
2. Dropped \`darwin/x86_64\` from \`required-environments\` (no Intel macOS wheels for numba/llvmlite in the lock).
3. Switched to \`license = "MIT"\` + \`license-files = ["LICENSE"]\` — confirmed no more deprecation warning on build.
4. CI now runs \`uv sync --locked --group dev\`.
5. Added \`.venv/\` to \`.gitignore\`.
6. Moved \`flake8\` and \`pytest\` from runtime \`dependencies\` into the \`dev\` dependency group.

\`uv.lock\` was regenerated (no changes to the resolved set, since the dev group already pulled these transitively).

bimac (approved):

PR #1134, opened 2026-08-23
Neuropixels synchronisation error detection

oliche:
## Summary
- `version3B` already sets `qc_all=False` when `_check_diff_3b`'s 150ppm burst check fails, but that signal was unlabeled and unenforced — a probe hitting a dropped-edge, duplicate-burst, or isolated-glitch pathology just silently fell back to `type='exact'` 
- New `classify_sync_anomaly()` labels which of the three it is, from the raw front times alone. Both `version3A`/`version3B` now log the label whenever their existing QC check fails, and accept a new opt-in `raise_on_anomaly=False` kwarg that raises a new `SyncFrontsAnomaly` exception instead of continuing silently.
- Default behavior is unchanged for every existing caller — this is detection + an opt-in exception, not a change to the sync algorithm, fit types, or 3A's frame2ttl/right_camera precedence.
- Found via a real audit of 20 IBL probes with flagged sync issues ([lfpack#8](https://github.com/int-brain-lab/lfpack/issues/8)) — 11 had one of these three defects (plus one `irregular_3A_reference` case, out of scope here).

## Test plan
- [x] New `ibllib/tests/test_sync_probes.py`: 5 unit tests for `classify_sync_anomaly` (clean + the three defect signatures), 5 wiring tests for `version3A`/`version3B`'s `raise_on_anomaly` (raises when `True`, doesn't when `False`/default) — all file I/O mocked, no external dataset needed.
- [x] `ibllib/tests/test_ephys.py` — unaffected (same pre-existing unrelated network-auth failure as before this change).
- [x] `ibllib/tests/integration/test_ephys_synchronization.py` — still just skips (no external dataset in this environment), confirmed it doesn't error on collection.
- [x] ruff clean.

### DIFF
--- a/ibllib/ephys/sync_probes.py
+++ b/ibllib/ephys/sync_probes.py
@@ -45,7 +45,7 @@ def sync(ses_path, **kwargs):
         return version3B(ses_path, **kwargs)
 
 
-def version3A(ses_path, display=True, type='smooth', tol=2.1, probe_names=None, raise_on_anomaly=False):
+def version3A(ses_path, display=True, type='smooth', tol=2.1, probe_names=None, raise_on_anomaly=True):
     """
     From a session path with _spikeglx_sync arrays extracted, locate ephys files for 3A and
      outputs one sync.timestamps.probeN.npy file per acquired probe. By convention the reference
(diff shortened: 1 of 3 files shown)


ibllib/ephys/sync_probes.py
```diff
@@ -110,11 +115,16 @@ def get_sync_fronts(auxiliary_name):
...
+                probe_name = ephys_file.path.parts[-1]
+                _logger.error(f'{ses_path} probe {probe_name}: sync anomaly using aux channel "{aux_used}"')
+                if raise_on_anomaly:
+                    raise SyncFrontsAnomaly(f'{ses_path} probe {probe_name}: aux channel "{aux_used}"')
         out_files = _save_timestamps_npy(ephys_file, timestamps, sr)
     return qc_all, out_files
 
 
-def version3B(ses_path, display=True, type=None, tol=2.5, probe_names=None):
+def version3B(ses_path, display=True, type=None, tol=2.5, probe_names=None, raise_on_anomaly=False):
```
oliche, line 131: Here the raise_on_anomaly should be True. 
Given what we saw, it is unlikely to happen on new recordings. But should this happen, raising is the lesser of the two evils of having this unnoticed in our database.

ibllib/ephys/sync_probes.py
```diff
@@ -45,14 +45,17 @@ def sync(ses_path, **kwargs):
...
+def version3A(ses_path, display=True, type='smooth', tol=2.1, probe_names=None, raise_on_anomaly=False):
     """
     From a session path with _spikeglx_sync arrays extracted, locate ephys files for 3A and
      outputs one sync.timestamps.probeN.npy file per acquired probe. By convention the reference
      probe is the one with the most synchronisation pulses.
      Assumes the _spikeglx_sync datasets are already extracted from binary data
     :param ses_path:
     :param type: linear, exact or smooth
+    :param raise_on_anomaly: if True, raise SyncFrontsAnomaly when the sync_probe_front_times
+     tolerance check fails for a probe, instead of only logging it and continuing with
```
oliche: Here add in the comment that this is for legacy extractions only, as no-one is using those probes anymore as of today (put the current date)

ibllib/exceptions.py
```diff
@@ -27,6 +27,14 @@ class Neuropixel3BSyncFrontsNonMatching(IblError):
     )
 
 
+class SyncFrontsAnomaly(IblError):
+    explanation = (
+        ' The raw sync pulse fronts show a specific, classifiable pathology: dropped edges, '
```
oliche, line 32: Maybe put a link to the original lfpack issue, where we'll describe those issues.

ibllib/ephys/sync_probes.py
grg2rsr: agreed

grg2rsr (commented):
looks good to me

grg2rsr (approved):

PR #1136, opened 2026-08-24
Migrate packaging to pyproject.toml - switch to uv build

bimac:
here's how to have several modules in a package.

please double-check if the included package data is correct (i.e., non-python files). See:
- https://setuptools.pypa.io/en/latest/userguide/datafiles.html
- https://docs.astral.sh/uv/concepts/build-backend/#file-inclusion-and-exclusion


oliche (approved):
I imported the old MANIFEST.ini

PR #1137, opened 2026-08-24
github actions overhaul

grg2rsr:
(the scope of this PR has expanded a bit, hence the rename, a bit out of sync now with the branch name)

 + `ruff check` has been part of the test suite, but not `ruff format`. Adding a check for it here that should fail if one tries to add unformatted code
+ the entire repo was re-ruffed, and the ruff commits added to the git blame ignore refs
+ Integration tests (on LightningAI) disabled for every push - only triggered by PRs against develop (or master) that are ready for review


.github/workflows/unit-tests.yml
```diff
@@ -39,6 +39,7 @@ jobs:
     - name: Flake8 & ruff
       run: |
         python -m ruff check .
+        python -m ruff format --check .
         python -m flake8 --select D --ignore E ibllib/qc/camera.py ibllib/qc/task_metrics.py
```
bimac: let's remove flake8 as discussed during stand-up, also from the dependencies.

we may also want to enable a few more rules in pyproject.toml (not necessarry as part of this PR). See [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-flake8).

bimac (changes requested):

.github/workflows/unit-tests.yml
grg2rsr: flake8 removed

.github/workflows/unit-tests.yml
```diff
@@ -8,8 +8,23 @@ on:
     branches: [ "*" ]
 
 jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
bimac: actions/checkout@v7 is available

.github/workflows/unit-tests.yml
```diff
@@ -8,8 +8,23 @@ on:
...
+      - uses: actions/setup-python@v5
```
bimac: actions/setup-python@v7 is available

.github/workflows/unit-tests.yml
```diff
@@ -25,21 +40,18 @@ jobs:
     env:
       ONE_SAVE_ON_DELETE: false
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v5
```
bimac: actions/setup-python@v7 is available

bimac (approved):
Abgesehen von den 4 Kommentaren alles gut.
Pass auf mit https://github.com/int-brain-lab/ibllib/pull/1132 - da gibt's evtl cross-talk

.github/workflows/unit-tests.yml
```diff
@@ -25,21 +40,18 @@ jobs:
     env:
       ONE_SAVE_ON_DELETE: false
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
```
bimac: actions/checkout@v7 is available

.github/workflows/unit-tests.yml
grg2rsr: done

.github/workflows/unit-tests.yml
grg2rsr: done

